PR #1, opened 2024-07-10
Ignore containers with invalid UUID in representation

fabiaserra:
## Description
When trying to save scenes that contain OpenPype UUIDs in representations Nuke was catching this error and not allowing you to save the scene:
```
  File "/sw/nuke/15.0v4/plugins/nuke_internal/callbacks.py", line 92, in onScriptSave
    _doCallbacks(onScriptSaves)
  File "/sw/nuke/15.0v4/plugins/nuke_internal/callbacks.py", line 46, in _doCallbacks
    f[0](*f[1],**f[2])
  File "/home/lmenga/.local/share/AYON/addons/nuke_0.2.4-ax.2/ayon_nuke/api/lib.py", line 844, in check_inventory_versions
    for repre_entity in repre_entities:
  File "/pipe/ayon/release/current/dependencies/ayon_api/server_api.py", line 5967, in get_representations
    for parsed_data in query.continuous_query(self):
  File "/pipe/ayon/release/current/dependencies/ayon_api/graphql.py", line 380, in continuous_query
    raise GraphQlQueryFailed(
ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Invalid entity ID 6500f58ef1d7617e88ffc758 on item 'project/representations' (Line 3 Column 5)
```
This PR fixes it by reusing existing functions that already address this issue.

## How to test
1. Create a scene and copy some nodes that contain OpenPype UUIDs
2. Save the scene

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -33,6 +33,7 @@ from ayon_core.addon import AddonsManager
 from ayon_core.pipeline.template_data import get_template_data_with_names
 from ayon_core.pipeline import (
     Anatomy,
+    registered_host,
     get_current_host_name,
     get_current_project_name,
     get_current_folder_path,
@@ -40,6 +41,7 @@ from ayon_core.pipeline import (
     AYON_INSTANCE_ID,
     AVALON_INSTANCE_ID,
 )
+from ayon_core.pipeline.load import filter_containers
 from ayon_core.pipeline.context_tools import (
     get_current_context_custom_workfile_template
 )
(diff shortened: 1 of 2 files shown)


client/ayon_nuke/api/lib.py
```diff
@@ -810,89 +812,24 @@ def check_inventory_versions():
...
-            continue
 
-        version_id = repre_entity["versionId"]
-        version_entity = version_entities_by_id.get(version_id)
-        if not version_entity:
-            log.warning((
-                "Could not find the version on node \"{}\""
-            ).format(node.name()))
-            continue
+    category_colors = {
```
fabiaserra, line 819: do we have any better place where we could define the different generic colors to use for these different categories?

I made `invalid` red and `not_found` as yellow... but previously we didn't do anything to color those, do we want to keep it that way?

BigRoy: You could move these to project settings if they feel like something that should be customizable for a studio. Otherwise as it is now or as contants should be fine. I'd keep it as is until a feature request to customize it comes in - that way, the implementation for now remains light-weight.

client/ayon_nuke/api/lib.py
```diff
@@ -810,89 +812,24 @@ def check_inventory_versions():
...
-        else:
-            color_value = "0xd84f20ff"
-        node["tile_color"].setValue(int(color_value, 16))
+    filtered_containers = filter_containers(containers, project_name)
+    for category, containers in filtered_containers._asdict().items():
+        container_color = category_colors.get(category) or category_colors["not_found"]
+        for container in containers:
+            container["node"]["tile_color"].setValue(
+                int(container_color, 16)
+            )
```
BigRoy, line 832: This may be nicer?
```suggestion
    for category, containers in filtered_containers._asdict().items():
        color = category_colors.get(category) or category_colors["not_found"]
        color = int(color, 16)  # convert hex to nuke tile color int
        for container in containers:
            container["node"]["tile_color"].setValue(color)
```

I'm just curious why you ended up doing the `or category_colors["not_found"]` fallback. It feels a bit as if it may be hiding another issue accidentally?

I think, if it's an unknown category for whatever reason we should just ignore it.

```suggestion
    for category, containers in filtered_containers._asdict().items():
        if category not in category_colors:
            continue
        color = category_colors[category]
        color = int(color, 16)  # convert hex to nuke tile color int
        for container in containers:
            container["node"]["tile_color"].setValue(color)
```

client/ayon_nuke/api/lib.py
fabiaserra: yeah I agree that is out of scope and should be a separate PR. Having that said, do you like the color palette chosen for the `not_found` and `invalid`?

client/ayon_nuke/api/lib.py
```diff
@@ -810,89 +812,24 @@ def check_inventory_versions():
     and check if the node is having actual version. If not then it will color
     it to red.
     """
```
BigRoy, line 814: Admittedly this docstring is weird, no? Should we update it with this PR?

```
    """Update loaded container nodes' colors based on version state.

    This will group containers by their version to outdated, not found,
    invalid or latest and colorize the nodes based on the category.
    """
```

client/ayon_nuke/api/lib.py
fabiaserra: > I'm just curious why you ended up doing the or category_colors["not_found"] fallback. It feels a bit as if it may be hiding another issue accidentally?

Yeah... just being cautious of possible changes on the API that could rephrase the categories or any other issues so it doesn't try to apply `None`. Happy to ignore those cases, perhaps even adding a warning?

client/ayon_nuke/api/lib.py
fabiaserra: I agree, updating it now

client/ayon_nuke/api/lib.py
BigRoy: > yeah I agree that is out of scope and should be a separate PR. Having that said, do you like the color palette chosen for the not_found and invalid?

Not a Nuke user myself ;) so can't test it visually inside of Nuke. I'll leave this to @jakubjezek001

BigRoy (approved):
Code looks good to me.

Initially I wasn't sure about the "private" `._asdict()` call to the `namedtuple` that `filter_containers` returns but I think there just isn't really a nicer way than that. So, fine with me.

Will leave the color choices and testing to someone else. :)

client/ayon_nuke/api/lib.py
jakubjezek001: I really like the approach! Thanks @fabiaserra. I wonder if we could just add the `category_colors` into **api/constants.py** for now.

jakubjezek001: Yeah, this is the best we are having now.

jakubjezek001 (approved):
@fabiaserra I've committed the suggested changes to your fork - thanks for allowing open commits. It's been tested with older OpenPype containers and everything works well.

Regarding what you're seeing in Manager, it shouldn't be an issue as long as it lets you work with the screen.
image

BigRoy (approved):
Don't have nuke on my end but code looks good!

fabiaserra:
> @fabiaserra I've committed the suggested changes to your fork - thanks for allowing open commits. It's been tested with older OpenPype containers and everything works well.

Awesome, thank you!!

PR #8, opened 2024-07-25
Repair action from validate correct asset context to fix the write node's filepath 

moonyuet:
## Changelog Description
This PR is to fix the bug of the validator for asset context not repairing the write nodes filepath as expected.

## Additional info
n/a

## Testing notes:

1. Launch Nuke
2. Change your folder path and task
3. Publish
4. Validate Correct Asset Context errors out
5. Click Repair
6. Should be fixing the filepath along with the correct context data.


client/ayon_nuke/api/lib.py
```diff
@@ -993,6 +993,7 @@ def format_anatomy(data):
         },
         "frame": "#" * frame_padding,
     })
+    print("data", data)
```
BigRoy, line 996:
```suggestion
```

client/ayon_nuke/plugins/publish/validate_asset_context.py
```diff
@@ -110,5 +112,29 @@ def repair(cls, instance):
...
+        })
+        anatomy_filled = format_anatomy(data)
+
+        # build file path to workfiles
+        fdir = str(
+            anatomy_filled["work"]["default"]["directory"]
+        ).replace("\\", "/")
+        data["work"] = fdir
+        fpath = StringTemplate(data["fpath_template"]).format_strict(data)
+        write_node["file"].setValue(fpath)
```
BigRoy, line 140: Shouldn't this be up to the Creator class instead to update the filepath if the context changes?

Fusion does that for example:
- https://github.com/ynput/ayon-fusion/blob/develop/client/ayon_fusion/api/plugin.py#L121-L138

That way it also works if you change the context via the publisher UI - not just in the validator. (And you can remove this logic from this validator then because it's already updating via the Creator above)

moonyuet: I added a separate function for updating instances and it works for saving changes but it doesn't work in repair action. It is mostly because we need to access the transient data to get the filepath and fix it? 
The repair action still needs `reset_write_nodes_filepath` to fix the filepath.

client/ayon_nuke/plugins/publish/validate_asset_context.py
```diff
@@ -111,4 +112,15 @@ def repair(cls, instance):
...
+            data = dict({
+                "folderPath": created_instance["folderPath"],
+                "task": created_instance["task"],
+                "productName": created_instance["productName"],
+                "productType": created_instance["productType"]
+            })
+
+            reset_write_node_filepath(instance_node, data)
+
+        create_context.save_changes()
```
BigRoy, line 126: This _should_ be redundant - because the update instances of the Creator should already handle it.
```suggestion
        create_context.save_changes()

```

If it doesn't - we should debug in `update_instances` why it does not trigger what we need there. We should be able to rely on the creator class to update the instance.

client/ayon_nuke/api/plugin.py
```diff
@@ -309,6 +310,39 @@ def set_selected_nodes(self, pre_create_data):
...
+                        created_inst["productName"]),
+                    "folderPath": changes["folderPath"].new_value or (
+                        created_inst["folderPath"]
+                    ),
+                    "task": changes["task"].new_value or (
+                        created_inst["task"]
+                    ),
+                    "productType": changes["productType"].new_value or (
+                        created_inst["productType"]
+                    )
```
BigRoy, line 330: This `changed_data` dict isn't actually the changed data.
This `x` or `y` logic doesn't need to be there if you just call: 

```python
data = created_inst.data_to_store()
keys = ("productName", "folderPath", "task", "productType")
context_values = {key: data[key] for key in keys}
reset_write_node_filepath(instance_node, changed_data)
```

However, I'd argue that even better would be if the `reset` logic was on the Creator and you'd just pass it the created instance so that the function call simplifies.

```python
def update_write_node_filepath(self, created_inst, changes):
    keys = ("productName", "folderPath", "task", "productType")
    if not any(key in changes for key in keys):
        # No relevant changes, no need to update
        return

    # Update values with new formatted path
    instance_node = created_inst.transient_data["node"]
    formatting_data = copy.deepcopy(data)
    write_node = nuke.allNodes(group=instance_node, filter="Write")[0]
    formatting_data.update({
        "fpath_template": (
        "{work}/renders/nuke/{subset}/{subset}.{frame}.{ext}"),
        "ext": write_node["file_type"].value()
    })
    anatomy_filled = format_anatomy(formatting_data)

    # build file path to workfiles
    fdir = str(
        anatomy_filled["work"]["default"]["directory"]
    ).replace("\\", "/")
    formatting_data["work"] = fdir
    fpath = StringTemplate(formatting_data["fpath_template"]).format_strict(
        formatting_data)
    write_node["file"].setValue(fpath)
    instance_node["name"].setValue(formatting_data["productName"])
```

The issue with the function in `lib` now is that it expects `data` but it is not clear, for a library function, what that data should need. It may be better to then turn it into passing the product name, etc. as separate arguments - but I don't think there's huge benefit now in moving it to `lib`.

BigRoy (changes requested):

client/ayon_nuke/plugins/publish/validate_asset_context.py
moonyuet: I have debugged and still dont understand why it's not working. It would be helpful if you can give some adivce.

client/ayon_nuke/api/plugin.py
```diff
@@ -222,10 +224,7 @@ def update_instances(self, update_list):
 
             # update instance node name if product name changed
             if "productName" in changes.changed_keys:
-                instance_node["name"].setValue(
-                    changes["productName"].new_value
-                )
-
+                update_write_node_filepath(created_inst, changes)
```
BigRoy, line 227: This feels like an odd change?

So this should ALSO change the filepath? If so, should it also do so on task change, etc? If so, we don't need this if statement here. and we can just always call it - and then also do not need to implement `update_instances` on the other plugin?

client/ayon_nuke/api/plugin.py
```diff
@@ -309,6 +311,26 @@ def set_selected_nodes(self, pre_create_data):
         else:
             self.selected_node = None
 
+    def update_instances(self, update_list):
+        for created_inst, changes in update_list:
+            instance_node = created_inst.transient_data["node"]
+
+            # update instance node name if product name changed
+            if "folderPath" or "task" in changes.changed_keys:
+                update_write_node_filepath(created_inst, changes)
```
BigRoy, line 320: No need to check - the function itself already checks, right?
Also with this check here it wouldn't update the name on product name changes.
```suggestion
            update_write_node_filepath(created_inst, changes)
```

client/ayon_nuke/api/plugin.py
```diff
@@ -1225,3 +1247,31 @@ def exposed_write_knobs(settings, plugin_name, instance_node):
         instance_node.addKnob(nuke.Text_Knob('', 'Write Knobs'))
     write_node = nuke.allNodes(group=instance_node, filter="Write")[0]
     link_knobs(exposed_knobs, write_node, instance_node)
+
+
+def update_write_node_filepath(created_inst, changes):
```
BigRoy, line 1252: Please add a docstring.
```suggestion
def update_write_node_filepath(created_inst, changes):
    """Update instance node on context changes.
    
    Whenever any of productName, folderPath, task or productType
    changes then update:
    - output filepath of the write node
    - instance node's name to the product name
    """
```

client/ayon_nuke/api/plugin.py
BigRoy: Or maybe it makes more sense to take the `productName` updating of that `name` knob on the instance node out of that `update_write_node_filepath` function... so that function DOES actually ONLY update the filepath, nothing else.

client/ayon_nuke/api/plugin.py
```diff
@@ -309,6 +311,24 @@ def set_selected_nodes(self, pre_create_data):
...
+            instance_node = created_inst.transient_data["node"]
+
+            update_write_node_filepath(created_inst, changes)
+            # in case node is not existing anymore (user erased it manually)
+            try:
+                instance_node.fullName()
+            except ValueError:
+                self.remove_instances([created_inst])
+                continue
+
```
BigRoy, line 325:
```suggestion
            # in case node is not existing anymore (user erased it manually)
            try:
                instance_node.fullName()
            except ValueError:
                self.remove_instances([created_inst])
                continue
            
            update_write_node_filepath(created_inst, changes)
```
I suppose we should first check for the deleted node?

client/ayon_nuke/api/plugin.py
```diff
@@ -1225,3 +1245,38 @@ def exposed_write_knobs(settings, plugin_name, instance_node):
...
+    # build file path to workfiles
+    fdir = str(
+        anatomy_filled["work"]["default"]["directory"]
+    ).replace("\\", "/")
+    formatting_data["work"] = fdir
+    fpath = StringTemplate(formatting_data["fpath_template"]).format_strict(
+        formatting_data)
+    write_node["file"].setValue(fpath)
+    instance_node["name"].setValue(formatting_data["productName"])
```
BigRoy, line 1282: Please confirm this "name" update occurs when you want it to @moonyuet - currently it will ALSO happen now when changing the folderPath, task or producType whereas previously it only did on change of product name?

moonyuet: This is more dependent on the task and productType, guess it would be better to leave it as the previous did.

client/ayon_nuke/api/plugin.py
```diff
@@ -309,6 +311,30 @@ def set_selected_nodes(self, pre_create_data):
...
+            if "productName" in changes.changed_keys:
+                instance_node["name"].setValue(
+                    changes["productName"].new_value
+                )
+            # in case node is not existing anymore (user erased it manually)
+            try:
+                instance_node.fullName()
+            except ValueError:
+                self.remove_instances([created_inst])
+                continue
```
BigRoy, line 328: This would also need to be moved after the node removal check because if it wouldn't exist I'm pretty sure this would error.
```suggestion
            # in case node is not existing anymore (user erased it manually)
            try:
                instance_node.fullName()
            except ValueError:
                self.remove_instances([created_inst])
                continue
            # update instance node name if product name changed
            if "productName" in changes.changed_keys:
                instance_node["name"].setValue(
                    changes["productName"].new_value
                )
```

BigRoy (approved):
This will need some Nuke testing by someone else. Maybe @jakubjezek001? 

But code changes seem ok.

PR #10, opened 2024-08-01
Display and View fallback options for settings and api

jakubjezek001:
After adding support for context-specific anatomy token resolution in color management display and view options, we must now ensure value validation and provide fallback options. We're introducing a semicolon separator in values, allowing for multiple options where the order is important.

## Additional informations
- colorspace methods were moved from lib to new nuke API module `colorspace.py`. Some of those function might need to move into core addon but for now lets keep them here. 

### Pre testing 
- version up dev number in package.py
- deploy addon version to your server and bundle
- make sure core colormanagement is enabled
- notice hints on `ayon+settings://nuke/imageio/viewer/display` 

### How to do DCC testing
- When conducting this testing, it is important to note that a published ocio config may not be present in your task context. Our objective is to determine the backward compatibility of the process and identify any potential issues.  It is possible for only certain shot folders to have published ocio configurations, while others may not. In such scenarios, a fallback option must be available. 
- In addition to this PR, the changes in https://github.com/ynput/ayon-core/pull/834 also need to be deployed. Once completed, the colorspace settings in the core addon should be adjusted to use the fallback option. The built-in **ACES 1.2** should be selected for this purpose.

1. To test baking targeting it would be good to disable clearing nodes in your code here **ayon-nuke\client\ayon_nuke\api\plugin.py:1090** - just add `#` on line start. Then after open Nuke. This way baking nodes created during **ExtractReviewIntermediates** processing will stay in script and you would be able to check how they are set. 

2. Follow next setps where we fork into two different testing
	a. _Baking Target Colorspace (Use Display & View)_
			- set `{project[code]}_{parent}_{folder[name]};sRGB` into View here **ayon+settings://nuke/imageio/baking_target/display_view/view**
			- Go to Nuke and publish render node
			- Check what is set into node just before write node.
			- it should be set to **sRGB** if ocio config is set to different then published product.
			
	b. _Baking Target Colorspace (Use Colorspace)_
			- set `{project[code]}_{parent}_{folder[name]};Output - Rec.709` into View here **ayon+settings://nuke/imageio/baking_target/colorspace**
			- Go to Nuke and publish render node
			- Check what is set into node just before write node.
			- it should be set to **Output - Rec.709** if ocio config is set to different then published product.

### DIFF
--- a/client/ayon_nuke/api/colorspace.py
+++ b/client/ayon_nuke/api/colorspace.py
@@ -259,26 +259,30 @@ def get_formatted_display_and_view_as_dict(
             display_views.append(
                 {"view": view.strip(), "display": display.strip()})
 
+    root_display_and_view = get_display_and_view_colorspaces(root_node)
     for dv_item in display_views:
         # format any template tokens used in the string
-        profile = {
-            "view": StringTemplate(dv_item["view"]).format_strict(
-                formatting_data),
-            "display": StringTemplate(dv_item["display"]).format_strict(
-                formatting_data) if dv_item["display"] else None,
-        }
-        log.debug("Resolved profile: `{}`".format(profile))
-
-        # making sure formatted colorspace exists in running session
-        # also need to test case where display is None
-        test_string = profile["view"]
-        if profile["display"]:
+        view = StringTemplate.format_strict_template(
+            dv_item["view"], formatting_data
+        )
+        test_string = view
+        display = dv_item["display"]
+        if display:
+            display = StringTemplate.format_strict_template(
+                display, formatting_data
+            )
             test_string = create_viewer_profile_string(
-                profile["view"], profile["display"], path_like=False
+                view, display, path_like=False
             )
 
-        if test_string in get_display_and_view_colorspaces(root_node):
-            return profile
+        log.debug(f"Resolved View: '{view}' Display: '{display}'")
+
+        # Make sure formatted colorspace exists in running session
+        if test_string in root_display_and_view:
+            return {
+                "view": view,
+                "display": display,
+            }
 
 
 def get_formatted_colorspace(
--- a/client/ayon_nuke/api/colorspace.py
+++ b/client/ayon_nuke/api/colorspace.py
@@ -265,6 +265,7 @@ def get_formatted_display_and_view_as_dict(
         view = StringTemplate.format_strict_template(
             dv_item["view"], formatting_data
         )
+        # for config without displays - nuke_default
         test_string = view
         display = dv_item["display"]
         if display:
@@ -277,7 +278,7 @@ def get_formatted_display_and_view_as_dict(
 
         log.debug(f"Resolved View: '{view}' Display: '{display}'")
 
-        # Make sure formatted colorspace exists in running session
+        # Make sure formatted colorspace exists in running ocio config session
         if test_string in root_display_and_view:
             return {
                 "view": view,
--- a/client/ayon_nuke/api/colorspace.py
+++ b/client/ayon_nuke/api/colorspace.py
@@ -21,7 +21,7 @@ DISPLAY_AND_VIEW_COLORSPACES_CACHE = {}
 COLORSPACES_CACHE = {}
 
 
-def get_display_and_view_colorspaces(root_node: nuke.Node) -> list:
+def get_display_and_view_colorspaces(root_node):
     """Get all possible display and view colorspaces
 
     This is stored in class variable to avoid multiple calls.
@@ -43,10 +43,7 @@ def get_display_and_view_colorspaces(root_node: nuke.Node) -> list:
     return DISPLAY_AND_VIEW_COLORSPACES_CACHE[script_name]
 
 
-def get_colorspace_list(
-    colorspace_knob: nuke.Knob,
-    node: nuke.Node = None,
-) -> list:
+def get_colorspace_list(colorspace_knob, node=None):
     """Get available colorspace profile names
 
     Args:
@@ -84,7 +81,7 @@ def get_colorspace_list(
     return COLORSPACES_CACHE[identifier_key]
 
 
-def colorspace_exists_on_node(node: nuke.Node, colorspace_name: str) -> bool:
+def colorspace_exists_on_node(node, colorspace_name):
     """ Check if colorspace exists on node
 
     Look through all options in the colorspace knob, and see if we have an
@@ -110,7 +107,7 @@ def colorspace_exists_on_node(node: nuke.Node, colorspace_name: str) -> bool:
     return colorspace_name in get_colorspace_list(colorspace_knob, node)
 
 
-def create_viewer_profile_string(viewer, display=None, path_like=False) -> str:
+def create_viewer_profile_string(viewer, display=None, path_like=False):
     """Convert viewer and display to string
 
     Args:
@@ -130,10 +127,7 @@ def create_viewer_profile_string(viewer, display=None, path_like=False) -> str:
 
 
 def get_formatted_display_and_view(
-    view_profile: dict,
-    formatting_data: dict,
-    root_node: nuke.Node = None,
-) -> str:
+        view_profile, formatting_data, root_node=None):
     """Format display and view profile into string.
 
     This method is formatting a display and view profile. It is iterating
@@ -209,10 +203,7 @@ def get_formatted_display_and_view(
 
 
 def get_formatted_display_and_view_as_dict(
-    view_profile: dict,
-    formatting_data: dict,
-    root_node: nuke.Node = None,
-) -> dict:
+        view_profile, formatting_data, root_node=None):
     """Format display and view profile into dict.
 
     This method is formatting a display and view profile. It is iterating
@@ -296,10 +287,7 @@ def get_formatted_display_and_view_as_dict(
 
 
 def get_formatted_colorspace(
-    colorspace_name: str,
-    formatting_data: dict,
-    root_node: nuke.Node = None,
-) -> str:
+        colorspace_name, formatting_data, root_node=None):
     """Format colorspace profile name into string.
 
     This method is formatting colorspace profile name. It is iterating
--- a/client/ayon_nuke/api/colorspace.py
+++ b/client/ayon_nuke/api/colorspace.py
@@ -16,8 +16,8 @@ import nuke
 log = Logger.get_logger(__name__)
 
 
-DISPLAY_AND_VIEW_COLORSPACES_CACHE = {}
-COLORSPACES_CACHE = {}
+_DISPLAY_AND_VIEW_COLORSPACES_CACHE = {}
+_COLORSPACES_CACHE = {}
 
 
 def get_display_and_view_colorspaces(root_node):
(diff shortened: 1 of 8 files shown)


client/ayon_nuke/api/lib.py
```diff
@@ -1545,19 +1547,66 @@ def set_viewers_colorspace(self, imageio_nuke):
...
+        views = [view_profile["view"]]
+        displays = [view_profile["display"]]
+
+        # separate all values by path separator
+        if COLOR_VALUE_SEPARATOR in view_profile["view"]:
+            views = view_profile["view"].split(
+                COLOR_VALUE_SEPARATOR)
+        if COLOR_VALUE_SEPARATOR in view_profile["display"]:
+            displays = view_profile["display"].split(
+                COLOR_VALUE_SEPARATOR)
```
BigRoy, line 1573:
```suggestion
        views = view_profile["view"].split(COLOR_VALUE_SEPARATOR)
        displays = view_profile["display"].split(COLOR_VALUE_SEPARATOR)
```
Isn't this the exact same thing?

client/ayon_nuke/api/lib.py
```diff
@@ -1545,19 +1547,66 @@ def set_viewers_colorspace(self, imageio_nuke):
...
+        for dv_item in display_views:
+            log.debug("Trying to set viewerProcess: `{}`".format(dv_item))
+            # format any template tokens used in the string
+            dv_item_resolved = StringTemplate(dv_item).format_strict(
+                self.formatting_data
+            )
+            log.info("Resolved viewerProcess: `{}`".format(dv_item_resolved))
+
+            if dv_item_resolved in self.get_display_and_view_colorspaces():
+                return dv_item_resolved
```
BigRoy, line 1594: Not sure how much views and displays there usually are. But it may be MUCH faster if you'd format the displays and views separately, because then you'd format `displays + views` amount of times, whereas now you are doing that `displays x views` amount of times.

moonyuet:
I tried with this setting below:
image
nuke launched with the display and view setting
image
The context works without issue.
image

moonyuet (approved):

client/ayon_nuke/api/lib.py
```diff
@@ -1545,19 +1547,66 @@ def set_viewers_colorspace(self, imageio_nuke):
...
+
+    def get_display_and_view_colorspaces(self):
+        """ Get all possible display and view colorspaces
+
+        This is stored in class variable to avoid multiple calls.
+
+        Returns:
+            list: all possible display and view colorspaces
+        """
+        if not self._display_and_view_colorspaces:
```
iLLiCiTiT, line 1604:
```suggestion
        if self._display_and_view_colorspaces is None:
```

client/ayon_nuke/api/lib.py
iLLiCiTiT, line 1594: I would do it in the loop.
```suggestion
        dv_colorspaces = self.get_display_and_view_colorspaces()
        for view in views:
            for display in displays:
                dv_item = create_viewer_profile_string(
                    view.strip(), display.strip(), path_like=False
                )

                log.debug("Trying to set viewerProcess: `{}`".format(dv_item))
                # format any template tokens used in the string
                dv_item_resolved = StringTemplate(dv_item).format_strict(
                    self.formatting_data
                )
                log.info(
                    "Resolved viewerProcess: `{}`".format(dv_item_resolved)
                )

                if dv_item_resolved in dv_colorspaces:
                    return dv_item_resolved
```


If the content of the loop should be separated then I would use submethod:
```python
        ...
        dv_colorspaces = self.get_display_and_view_colorspaces()
        for view in views:
            for display in displays:
                dv_item_resolved = self._get_resolved_display_view_item(
                    view, display, dv_colorspaces
                )
                if dv_item_resolved:
                    return dv_item_resolved

    def _get_resolved_display_view_item(
        self, view, display, available_colorspaces
    ):
        dv_item = create_viewer_profile_string(
            view.strip(), display.strip(), path_like=False
        )

        log.debug("Trying to set viewerProcess: `{}`".format(dv_item))
        # format any template tokens used in the string
        dv_item_resolved = StringTemplate(dv_item).format_strict(
            self.formatting_data
        )
        log.info(
            "Resolved viewerProcess: `{}`".format(dv_item_resolved)
        )

        if dv_item_resolved in available_colorspaces:
            return dv_item_resolved
        return None
```

jakubjezek001: This will go only 2 max 3 times, so I would say this will not be an issue.

jakubjezek001: Thank you for the suggestions. I was refactoring the code a little and this wouldn't apply anymore. Thanks again!

jakubjezek001:
> I tried with this setting below

@moonyuet thank you very much for the testing. Yeah that is looking promissing. I wonder if you could rather use anatomy tokens, something like `{project[code]}_{parent}_{folder[name]};sRGB` 
image

This way we should be able to even test that formating on context is working.

moonyuet:
> > I tried with this setting below
> 
> @moonyuet thank you very much for the testing. Yeah that is looking promissing. I wonder if you could rather use anatomy tokens, something like `{project[code]}_{parent}_{folder[name]};sRGB` image
> 
> This way we should be able to even test that formating on context is working.

It's working
image
image

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+
+from .constants import COLOR_VALUE_SEPARATOR
+
+import nuke
+
+log = Logger.get_logger(__name__)
+
+
+DISPLAY_AND_VIEW_COLORSPACES_CACHE = {}
+COLORSPACES_CACHE = {}
```
iLLiCiTiT, line 20: Should these be imported? If not mark them with underscore...
```suggestion
_DISPLAY_AND_VIEW_COLORSPACES_CACHE = {}
_COLORSPACES_CACHE = {}
```

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+
+    This is stored in class variable to avoid multiple calls.
+
+    Args:
+        root_node (nuke.Node): root node
+
+    Returns:
+        list: all possible display and view colorspaces
+    """
+    global DISPLAY_AND_VIEW_COLORSPACES_CACHE
```
iLLiCiTiT, line 34: You don't need to use `global` when working with object.
```suggestion
```

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+    results = []
+
+    # making sure any node is provided
+    node = node or nuke.root()
+    # unique script based identifier
+    script_name = nuke.root().name()
+    node_name = node.fullName()
+    identifier_key = f"{script_name}_{node_name}"
+
+    if not COLORSPACES_CACHE.get(identifier_key):
```
iLLiCiTiT, line 64:
```suggestion
    if COLORSPACES_CACHE.get(identifier_key) is None:
```

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+    # generate all possible combination of display/view
+    display_views = []
+    for view in views:
+
+        if displays:
+            for display in displays:
+                display_views.append(
+                    {"view": view.strip(), "display": display.strip()})
+        else:
+            display_views.append({"view": view.strip(), "display": None})
```
iLLiCiTiT, line 264: This would make it more readable
```suggestion
        if not displays:
            display_views.append({"view": view.strip(), "display": None})
            contine

        for display in displays:
            display_views.append(
                {"view": view.strip(), "display": display.strip()}
            )
```

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+        root_node = nuke.root()
+
+    views = view_profile["view"].split(COLOR_VALUE_SEPARATOR)
+    displays = view_profile["display"].split(COLOR_VALUE_SEPARATOR)
+
+    # display could be optional in case nuke_default ocio config is used
+    if view_profile["display"]:
+        displays = view_profile["display"].split(COLOR_VALUE_SEPARATOR)
+    else:
+        displays = []
```
iLLiCiTiT, line 253: Double split of `"display"`.
```suggestion

    # display could be optional in case nuke_default ocio config is used
    displays = []
    if view_profile["display"]:
        displays = view_profile["display"].split(COLOR_VALUE_SEPARATOR)
```

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,338 @@
...
+                    {"view": view.strip(), "display": display.strip()})
+        else:
+            display_views.append({"view": view.strip(), "display": None})
+
+    for dv_item in display_views:
+        # format any template tokens used in the string
+        profile = {
+            "view": StringTemplate(dv_item["view"]).format_strict(
+                formatting_data),
+            "display": StringTemplate(dv_item["display"]).format_strict(
```
iLLiCiTiT, line 271: `dv_item["display"]` can be `None`, so I guess this would crash in that case?

jakubjezek001: Following is the full code and I believe this should not crash since if display key value is None it will set the value to None and skip the StringTemplate processing:

```
{
	"display": StringTemplate(dv_item["display"]).format_strict(
    	formatting_data) if dv_item["display"] else None,
 }
```

moonyuet (approved):
Tested with baking colorspace looks good
image
I even tested with other parameters too:
image

And the view transform one(I tried with different parameters):
image

So far so good.

client/ayon_nuke/api/colorspace.py
iLLiCiTiT: Nope, this line https://github.com/ynput/ayon-nuke/pull/10/files#diff-2198068d7d5b277a506dc19d3f56005a6ffbc780a4156a07676f32d3676e8b4dR255 will add `{"view": view.strip(), "display": None}` to `display_views`, so `"display"` here can be `None`.

client/ayon_nuke/api/colorspace.py
```diff
@@ -0,0 +1,334 @@
...
+        # making sure formatted colorspace exists in running session
+        # also need to test case where display is None
+        test_string = profile["view"]
+        if profile["display"]:
+            test_string = create_viewer_profile_string(
+                profile["view"], profile["display"], path_like=False
+            )
+
+        if test_string in get_display_and_view_colorspaces(root_node):
+            return profile
```
iLLiCiTiT, line 281:
```suggestion
    root_display_and_view = get_display_and_view_colorspaces(root_node)
    for dv_item in display_views:
        # format any template tokens used in the string
        view = StringTemplate.format_strict_template(
            dv_item["view"], formatting_data
        )
        display = dv_item["display"]
        if display:
            display = StringTemplate.format_strict_template(
                display, formatting_data
            )

        log.debug(f"Resolved View: '{view}' Display: '{display}'")

        # making sure formatted colorspace exists in running session
        # also need to test case where display is None
        test_string = view
        if display:
            test_string = create_viewer_profile_string(
                view, display, path_like=False
            )

        if test_string in root_display_and_view:
            return {
                "view": view,
                "display": display,
            }
```

BigRoy: `# Make sure formatted colorspace exists in running session` is that comment now in the wrong place?

iLLiCiTiT: > # Make sure formatted colorspace exists in running session is that comment now in the wrong place?

I would say it is related to `if test_string in root_display_and_view:` condition.

moonyuet (approved):
Looks good
Viewer
image
Bake Colorspace(Use Colorspace)
image
Bake Colorspace(Display & View)
image

PR #13, opened 2024-08-26
Use task attributes for WorkfileSettings

BigRoy:
## Changelog Description

Workfile resolution and FPS is now set using the current task's entity attribute instead of from the parent folder entity.

## Additional info

I didn't test it.

## Testing notes:

Create a folder with a task where the task has different FPS and resolution.

1. Create new workfile - settings should adhere to task attributes
2. Use "Reset resolution" and "Reset frame range" menu actions - it should set it using the task attributes
3. Publishing validations should also adhere to the task attributes, not the folder entity attributes. (If there are any validators for that.)

### DIFF
--- a/client/ayon_nuke/plugins/publish/validate_script_attributes.py
+++ b/client/ayon_nuke/plugins/publish/validate_script_attributes.py
@@ -34,7 +34,7 @@ class ValidateScriptAttributes(
         # Task may be optional for an instance
         task_entity = instance.data.get("taskEntity")
         if task_entity:
-            src_attributes = task_entity["attributes"]
+            src_attributes = task_entity["attrib"]
         else:
             src_attributes = instance.data["folderEntity"]["attrib"]
 
--- a/client/ayon_nuke/plugins/publish/validate_script_attributes.py
+++ b/client/ayon_nuke/plugins/publish/validate_script_attributes.py
@@ -31,7 +31,12 @@ class ValidateScriptAttributes(
 
         script_data = deepcopy(instance.context.data["scriptData"])
 
-        src_folder_attributes = instance.data["folderEntity"]["attrib"]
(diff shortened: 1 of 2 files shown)


moonyuet (changes requested):
When I tried to publish the workfile, it errors out. 
```
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2402141620_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\ayon-addon_template\ayon-nuke\client\ayon_nuke\plugins\publish\validate_script_attributes.py", line 37, in process
    src_attributes = task_entity["attributes"]
KeyError: 'attributes'

```

client/ayon_nuke/plugins/publish/validate_script_attributes.py
```diff
@@ -31,7 +31,12 @@ def process(self, instance):
 
         script_data = deepcopy(instance.context.data["scriptData"])
 
-        src_folder_attributes = instance.data["folderEntity"]["attrib"]
+        # Task may be optional for an instance
+        task_entity = instance.data.get("taskEntity")
+        if task_entity:
+            src_attributes = task_entity["attributes"]
```
moonyuet, line 37: it is a dictionary or no?
```suggestion
            src_attributes = task_entity.get("attributes", {})
```

BigRoy: Sorry, this was my bad. Typo.

Fixed with https://github.com/ynput/ayon-nuke/pull/13/commits/c5541d1f81c9fc89b3952c2743bad534b8175dda

moonyuet (approved):
The validator looks good now. And the task frame range being set correctly too.
image

PR #15, opened 2024-09-02
Implement Extract Nuke camera as FBX file.

robin-ynput:
## Changelog Description

This PR addresses: AY-6599
* Add support to extract camera as FBX file
* Add a new package server setting for a switch between ABC and FBX
* Ensure loader support generated FBX properly

## Additional info
Tested with Nuke 15.1v2

## Testing notes:

1. Build ayon-nuke package and upload it to server
2. Ensure Export to FBX is set in newly added `Nuke -> Publish Plugins -> Extract Camera Format` presets
image

3. Create and Publish a new camera from Nuke
4. Ensure resulting camera can be properly loaded back

### DIFF
--- a/client/ayon_nuke/plugins/load/load_camera.py
+++ b/client/ayon_nuke/plugins/load/load_camera.py
@@ -15,18 +15,17 @@ from ayon_nuke.api.lib import (
 )
 
 
-class AlembicCameraLoader(load.LoaderPlugin):
+class _NukeCameraLoader(load.LoaderPlugin):
     """
-    This will load alembic camera into script.
+    This will load a camera into script.
     """
 
     product_types = {"camera"}
     representations = {"*"}
-    extensions = {"abc"}
+    enabled = False
 
     settings_category = "nuke"
 
-    label = "Load Alembic Camera"
     icon = "camera"
     color = "orange"
     node_color = "0x3469ffff"
@@ -57,12 +56,21 @@ class AlembicCameraLoader(load.LoaderPlugin):
         file = self.filepath_from_context(context).replace("\\", "/")
 
         with maintained_selection():
-            camera_node = nuke.createNode(
-                "Camera2",
-                "name {} file {} read_from_file True".format(
-                    object_name, file),
-                inpanel=False
-            )
+
+            try:
+                camera_node = nuke.createNode(
+                    "Camera3",
+                    "name {} file {} read_from_file True".format(
+                        object_name, file),
+                    inpanel=False
+                )
+            except RuntimeError: # older nuke version
+                camera_node = nuke.createNode(
+                    "Camera2",
+                    "name {} file {} read_from_file True".format(
+                        object_name, file),
+                    inpanel=False
+                )                
 
             camera_node.forceValidate()
             camera_node["frame_rate"].setValue(float(fps))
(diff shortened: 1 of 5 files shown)


client/ayon_nuke/plugins/load/load_camera.py
```diff
@@ -196,3 +204,52 @@ def remove(self, container):
...
+            # WARNING - Nuke 15: Camera3 validate does not play along
+            # very well when mixing abc + fbx nodes in the same script.
+            # They need to reloaded manually.
+            #
+            # Hopefully this will be fixed by upcoming Camera4            
+            raise RuntimeError(
+                "Could not load incoming FBX camera properly. "
+                "This could be caused by a mix of abc and fbx "
+                "into the script."
+            )
```
robin-ynput, line 250: You can reproduce this by trying to load a mix of Camera from published ABC and FBX files:
1. Load an ABC camera
2. Append a FBX camera to the same Nuke scene
3. FBX Camera node is not properly loaded 

After some testing I feel this is coming from Nuke trying to be smart and re-using the **fbx_node_name** knob.

for Alembic
image


for FBX
image

MilaKudr (approved):

client/ayon_nuke/plugins/load/load_camera.py
```diff
@@ -196,3 +204,52 @@ def remove(self, container):
...
             nuke.delete(node)
+
+
+class AlembicCameraLoader(_NukeCameraLoader):
+    """
+    This will load alembic camera into script.
+    """
+    extensions = {"abc"}
+    label = "Load Alembic Camera"
+    enabled = True
```
moonyuet, line 215:
```suggestion
class AlembicCameraLoader(_NukeCameraLoader):
    """
    This will load alembic camera into script.
    """
    extensions = {"abc"}
    label = "Load Alembic Camera"
    enabled = True
```
I dont think it's necessary to create AlembicCameraLoader inherited from _NukeCameraLoader, we can just let the AlembicCameraLoader becoming the base class of the camera loader. Something like:
```python
class CameraLoader(load.LoaderPlugin):
    """
    This will load alembic camera into script.
    """
    extensions = {"abc"}
    label = "Load Alembic Camera"
    enabled = True

    def load(self, context, name, namespace, data):
        # your code

class FbxCameraLoader(CameraLoader):
    """
    This will load fbx camera into script.
    """
    extensions = {"fbx"}
    label = "Load FBX Camera"

    def load(self, context, name, namespace, data):
        # your code
```

client/ayon_nuke/plugins/publish/extract_camera.py
```diff
@@ -19,14 +20,39 @@ class ExtractCamera(publish.Extractor):
...
-        ("writePointClouds", False),
-        ("writeAxes", False)
-    ]
+    def _get_camera_export_presets(self, instance):
+        """
+        Args:
+            instance (dict): The current instance being published.
+
+        Returns:
+            presets (list.): The camera export presets to use.
```
moonyuet, line 29:
```suggestion
            list: The camera export presets to use.
```

client/ayon_nuke/plugins/load/load_camera.py
moonyuet: @jakubjezek001, want to hear your thought on this. 
Do you think that it's good to keep the code this way or we can optimize the code as I suggested?

BigRoy: From my POV - I understand @robin-ynput's point - however, if we go down that route please move out that abstract class from the file, move it somewhere into another file and do not import it into this file but import the file that contains it.

✅ 
```python
from x.y import z

class Plugin(z.CameraLoader)
```
Instead of:

🔴 
```python
from x.y.z import CameraLoader

class Plugin(CameraLoader)
```

Just so that `pyblish` plugin discovery does not even find that class and ignores it completely. It's just cleaner from the perspective of Pyblish not loading it (and hence it also not appearing in the "Skipped Plugins" list in publisher UI, etc.)

I'm fine with either way of the proposed ways - but just wanted to note this.

robin-ynput: That makes perfect sense. 
In that case I'll go ahead and implement initial @moonyuet suggestion and merge this one, just to ensure this goes out today. (if someone want to change that after, I'll do the work no problem).

PR #18, opened 2024-09-13
disabled raw data on clip when setting colorspace

ReeceMulley:
## Changelog Description
When loading a clip with a colorspace set, ie through File Rules, _Raw Data_ should be disabled on the read nodes. Just saves one click and a potential artist error

## Additional info
first PR!

## Testing notes:

1. Configure a file rule following the official Ayon core addon docs
2. Load in a clip. I used an exr of product type plate
3. Check "raw" knob

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -422,6 +422,7 @@ class LoadClip(plugin.NukeLoader):
         ):
             self.log.info(f"Used colorspace: {used_colorspace}")
             read_node["colorspace"].setValue(used_colorspace)
+            read_node["raw"].setValue(False)
         else:
             self.log.info("Colorspace not set...")
 


client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -422,6 +422,7 @@ def set_colorspace_to_node(
         ):
             self.log.info(f"Used colorspace: {used_colorspace}")
             read_node["colorspace"].setValue(used_colorspace)
+            read_node["raw"].setValue(False)
```
jakubjezek001, line 425: Since RAW is by default set to `False` it seems that your `init.py` somewhere might be overriding it. Please check your configuration.

jakubjezek001 (changes requested):
It seems redundant to add the value, please check your configuration for potential installation default overrides.

client/ayon_nuke/plugins/load/load_clip.py
ReeceMulley: you were absolutely right! Not sure how I missed it. Will close the PR

PR #19, opened 2024-09-13
AY-971 Refactor staging directory handling in Nuke creators

jakubjezek001:
# Change description
resolve https://github.com/ynput/ayon-nuke/issues/21

Implementing custom staging directory for all of Nuke products.

Bugfixes:
* `scriptmenus` still loading `openpype` modules
* selection validation code for all creators
* update/remove on instance from manually deleted nodes
* temporary render path settings ignored 

# Additional informations
- implementing changes from https://github.com/ynput/ayon-core/pull/887

# Testing notes

1. Enable custom staging directory with some profiles `ayon+settings://core/tools/publish/custom_staging_dir_profiles` (make it persistent so you can troubleshoot)
2. Create a template matching the staging directory under `ayon+anatomy://Dummy/templates/staging`
e.g. `{root[work]}/{project[name]}/{hierarchy}/{folder[name]}/temp_render_only/{product[type]}/{product[name]}`

3. Open Nuke and create+publish some products
4. Ensure that the staging directory is working is properly redirected to the relevant template for the associated products/task


robin-ynput:
Do you need review on this one or is it still draft ?
Sorry just asking cause I might lack some context and guidance on what to test to approve this one.

client/ayon_nuke/api/plugin.py
```diff
@@ -300,55 +310,76 @@ def integrate_links(self, node, outputs=True):
...
         else:
+            self.selected_nodes = []
             self.selected_node = None
 
     def update_instances(self, update_list):
+        super().update_instances(update_list)
         for created_inst, changes in update_list:
-            instance_node = created_inst.transient_data["node"]
+            if self.create_context.get_instance_by_id(created_inst.id):  # ensure was not deleted by super()
+                self._update_write_node_filepath(created_inst, changes)
```
iLLiCiTiT, line 330:
```suggestion
            # ensure was not deleted by super()
            if self.create_context.get_instance_by_id(created_inst.id):  
                self._update_write_node_filepath(created_inst, changes)
```

client/ayon_nuke/plugins/publish/extract_camera.py
```diff
@@ -53,8 +53,15 @@ def _get_camera_export_presets(self, instance):
...
-
     def process(self, instance):
+
+        # pass staging dir data
+        if not instance.data.get("stagingDir"):
+            staging_dir = os.path.normpath(
+                os.path.dirname(instance.data["path"]))
+            instance.data["stagingDir"] = staging_dir
+
+        staging_dir = instance.data["stagingDir"]
```
iLLiCiTiT, line 64:
```suggestion
        staging_dir = instance.data.get("stagingDir")
        if not staging_dir:
            staging_dir = os.path.normpath(
                os.path.dirname(instance.data["path"]))
            instance.data["stagingDir"] = staging_dir
        
```

client/ayon_nuke/plugins/publish/extract_review_intermediates.py
```diff
@@ -42,10 +42,10 @@ def process(self, instance):
...
             instance.data["representations"] = []
 
-        staging_dir = os.path.normpath(
-            os.path.dirname(instance.data["path"]))
-
-        instance.data["stagingDir"] = staging_dir
+        if not instance.data.get("stagingDir"):
+            staging_dir = os.path.normpath(
+                os.path.dirname(instance.data["path"]))
+            instance.data["stagingDir"] = staging_dir
```
iLLiCiTiT, line 48:
```suggestion
            instance.data["stagingDir"] = os.path.normpath(
                os.path.dirname(instance.data["path"]))
```

client/ayon_nuke/api/plugin.py
```diff
@@ -269,6 +276,9 @@ class NukeWriteCreator(NukeCreator):
     product_type = "write"
     icon = "sign-out"
 
+    temp_rendering_path_template = (  # default to be applied is settings is missing
+        "{work}/renders/nuke/{subset}/{subset}.{frame}.{ext}")
```
iLLiCiTiT, line 280:
```suggestion
        "{work}/renders/nuke/{product[name]}/{product[name]}.{frame}.{ext}")    
```

client/ayon_nuke/plugins/create/create_source.py
```diff
@@ -80,11 +82,18 @@ def create(self, product_name, instance_data, pre_create_data):
...
+
+        Raises:
+            NukeCreatorError. When provided selection is invalid.
+        """
+        if not pre_create_data.get("use_selection"):
+            raise NukeCreatorError(
+                "Creator error: only supported with active selection"
+            )
+
+        super()._set_selected_nodes(pre_create_data)
```
iLLiCiTiT, line 99:
```suggestion
        super()._set_selected_nodes(pre_create_data)

```

client/ayon_nuke/api/plugin.py
```diff
@@ -73,18 +74,21 @@ class NukeCreatorError(CreatorError):
...
+    def _pass_pre_attributes_to_instance(
         self,
         instance_data,
         pre_create_data,
         keys=None
     ):
-        if not keys:
-            keys = pre_create_data.keys()
-
+        keys = keys or pre_create_data.keys()
```
iLLiCiTiT, line 83: Maybe we should allow empty `keys`?
```suggestion
        if keys is None:
            keys = pre_create_data.keys()
```

client/ayon_nuke/api/plugin.py
```diff
@@ -158,7 +162,7 @@ def create_instance_node(
 
         return created_node
 
-    def set_selected_nodes(self, pre_create_data):
+    def _set_selected_nodes(self, pre_create_data):
```
iLLiCiTiT, line 165: I have a question, why we do store selected nodes to the object of create plugin? Wouldn't be better to have helper methods that would just return then?

```

Comment (robin-ynput) [reply client/ayon_nuke/api/plugin.py:166]: So I gave a try to this approach as indeed it make things simpler, but actually the necessity to store selected file raised immediatly. A lot of the creators manipulate the node graph during their instance(s) creation and this lost the intial selected nodes.
I'm adding a comment to reflect this.

Comment (iLLiCiTiT) [reply client/ayon_nuke/api/plugin.py:166]: Well, shouldn't that be called once in create plugin (at the start), and then pass the node where it is needed. So we don't rely on `self.selected_node`, which might be changed in some method -> it is really hard to find where it is really set and keep track of what is stored there at the moment it is used.

File: client/ayon_nuke/api/plugin.py [hunk 8]
```diff
@@ -73,18 +74,22 @@ class NukeCreatorError(CreatorError):
...
         creator_attrs = instance_data["creator_attributes"] = {}
-        for pass_key in keys:
-            creator_attrs[pass_key] = pre_create_data[pass_key]
+
+        creator_dict = {
+            key: value
+            for key, value in pre_create_data.items()
+            if key in keys
+        }
+        creator_attrs.update(creator_dict)
```
iLLiCiTiT, line 92:
```suggestion
        creator_attrs.update({
            key: value
            for key, value in pre_create_data.items()
            if key in keys
        })
```

client/ayon_nuke/api/plugin.py
robin-ynput: I reworked the node selection logic so we no more rely on the `self.selected_node` class properties. 
Please let me know if that corresponds better to what you had in mind.

client/ayon_nuke/api/plugin.py
```diff
@@ -158,26 +165,54 @@ def create_instance_node(
...
-    def set_selected_nodes(self, pre_create_data):
-        if pre_create_data.get("use_selection"):
-            self.selected_nodes = nuke.selectedNodes()
-            if self.selected_nodes == []:
-                raise NukeCreatorError("Creator error: No active selection")
+    def _get_current_selected_nodes(
+            self,
+            pre_create_data,
+            class_name: str = None,
+        ):
```
iLLiCiTiT, line 172:
```suggestion
        self,
        pre_create_data,
        class_name: str = None,
    ):
```

client/ayon_nuke/api/plugin.py
```diff
@@ -158,26 +165,54 @@ def create_instance_node(
...
+
+        if class_name:
+            selected_nodes = [
+                node
+                for node in selected_nodes
+                if node.Class() == class_name
+            ]
+
+        if class_name and use_selection and not selected_nodes:
+            raise NukeCreatorError(f"Creator error: Select a {class_name} node.")
```
iLLiCiTiT, line 198: Why the message says `Creator error:`?
```suggestion
            raise NukeCreatorError(f"Select a {class_name} node.")
```

client/ayon_nuke/api/plugin.py
```diff
@@ -304,55 +346,91 @@ def integrate_links(self, node, outputs=True):
...
-            elif len(selected_nodes) > 1:
-                NukeCreatorError("Creator error: Select only one camera node")
-            self.selected_node = selected_nodes[0]
-        else:
-            self.selected_node = None
+    def _get_current_selected_nodes(
+            self,
+            pre_create_data,
+            class_name: str = None,
+        ):
```
iLLiCiTiT, line 353:
```suggestion
        self,
        pre_create_data,
        class_name: str = None,
    ):
```

client/ayon_nuke/api/plugin.py
robin-ynput: thanks for this.

client/ayon_nuke/api/plugin.py
robin-ynput: Not sure, it was there initially so I've kept it. But I think you're right we can remove it.

client/ayon_nuke/api/plugin.py
robin-ynput: Yeah here as well, that's how things were implemented initially so I didn't really question it to be honest. 
I'll move the check out of this function to keep things consistent.

iLLiCiTiT (approved):

iLLiCiTiT:
Requires to bump required `core` version in package.py (>=1.0.12).

client/ayon_nuke/plugins/create/create_write_render.py
```diff
@@ -58,20 +38,22 @@ def create_instance_node(self, product_name, instance_data):
         write_data.update(instance_data)
 
         # get width and height
-        if self.selected_node:
+        if node_selection:  # contains 1 Write node or nothing
```
jakubjezek001, line 41: The selection seems to be broken now. Try it @robin-ynput following way:
1. select read node
2. open creator
3. chose Render (write)
4. keep Use selection activated
5. Create
Notice that popup will with empty window will tell you Wtite node....

jakubjezek001 (commented):
It was working as expected from test notes. Tested render, prerender, image nodes were correctly adding even the path from template inluding version token. Version was estimated next available version. 

I was testing backward compatibility and it also works as expected. 

There is only the mentioned issue with Creating and selection Popup.

I am not able to approve it since I am the original author of the PR but it does have Green light from me and is ready to merge.

client/ayon_nuke/plugins/create/create_write_render.py
robin-ynput: Thanks a lot, yep I realized I changed the previous behavior. 
I have fixed that.

Issue #33, opened 2024-10-18
AYON 'onScriptSave' callback can block saving workfile if it errors


jrsndl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Ayon uses onScriptSave callback to update nodes' colors. If this fails by any reason, artist is not able to save the file.

https://github.com/ynput/ayon-nuke/blob/b92271160f3547f986e0b9425275ef7f11b2bbdf/client/ayon_nuke/api/pipeline.py#L166

### Expected Behavior:

Failing the callback should not block the save.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Open Nuke workfile, make some changes
2. Stop Ayon server
3. Save

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

BigRoy:
Thanks for the report.

1. Do you have a reproducable how to make it fail? 
    - And does it still log an error / provide a stack trace? If so, could you attach one?
2. Does it work if we make the callback itself run completely within a `try..except`

jrsndl:
> 1. Do you have a reproducable how to make it fail?

This did happen when artist lost connection to the Ayon server, I will have to find a way to sabotage it.

>    * And does it still log an error / provide a stack trace? If so, could you attach one?

Yes. I didn't save the trace, but it failed here:
[    filtered_containers = filter_containers(containers, project_name)
](https://github.com/ynput/ayon-nuke/blob/b92271160f3547f986e0b9425275ef7f11b2bbdf/client/ayon_nuke/api/lib.py#L822)

> 2. Does it work if we make the callback itself run completely within a `try..except`

Yes, try / except in check_inventory_versions worked.

Sorry fon non-complete report, I had to help artists yesterday, making issue today so I do not forget.

BTW I had the same problem in OpenPype, once in a while the OpenPype data in Nuke node get corrupted, and artist was not able to save.

PR #36, opened 2024-10-23
Nuke behavior more resilient

jrsndl:
## Changelog Description
Ayon uses onScriptSave callback to update nodes' colors. If this fails by any reason, artist is not able to save the file.

Nuke sometimes fails on getting knob type needed in ```read_avalon_data```. This can result in artisn not being able to open workfile. Reason unknown, hard to test.

Changes allow artist to continue working.

## Additional info
Might close https://github.com/ynput/ayon-nuke/issues/33

## Testing notes:
1. Launch Nuke, open Workfile
2. Make changes
3. Stop server
4. Save file with Control + S

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -588,15 +588,18 @@ def read_avalon_data(node):
     if result:
         first_user_knob = result.group(2)
         # Collect user knobs from the end of the knob list
+        node_name = node['name'].value()
         for knob in reversed(node.allKnobs()):
             knob_name = knob.name()
             if not knob_name:
                 # Ignore unnamed knob
                 continue
-
-            knob_type = nuke.knob(knob.fullyQualifiedName(), type=True)
-            value = knob.value()
-
+            try:
+                knob_type = nuke.knob(knob.fullyQualifiedName(), type=True)
+                value = knob.value()
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/api/lib.py
```diff
@@ -593,10 +593,13 @@ def compat_prefixed(knob_name):
...
-            knob_type = nuke.knob(knob.fullyQualifiedName(), type=True)
-            value = knob.value()
-
+            try:
+                knob_type = nuke.knob(knob.fullyQualifiedName(), type=True)
+                value = knob.value()
+            except Exception:
+                log.debug(
+                    f"Error in knob {knob_name}, node {node['name'].value()}")
+                continue
```
jakubjezek001, line 602: This is a bit of a puzzle to be honest. I am not sure how this was blocking the processing. The code is just getting knob type and value. This seems to be redundant.

jrsndl: Nuke sometimes breaks on the knob type line. This can block opening Nuke scripts.

jrsndl: Unfortunatelly, I can't find any examples stached for this; it only happens few times in few thousands shots.

jakubjezek001 (approved):
looking good

PR #43, opened 2024-11-22
Easier Addon Setting for Nodes settings in color management section 

### LINKED ISSUE #42: Easier Addon Setting for Nodes settings in color management section

Nodes settings in color management section expects the admins to know the class of the creator plugins. `ayon+settings://nuke/imageio/nodes`

IMO, this is hard. 

### Suggested implementation?

Reuse the names/labels that admins/nuke artists are familiar with. 

where the setting shows a drop down menu and select the creators 

| Publisher | Settings suggestion |
|--|--|
| Image |  Image |

<details><summary>To achieve the result in my screenshot</summary>

I've replaced [these lines](https://github.com/ynput/ayon-nuke/blob/2c8af78bebad13070adae7317e315a7c8bb2f105/server/settings/imageio.py#L15-L23) with 

```python
def nuke_creator_plugins_enum():
    return [
        {"value": "CreateWritePrerender", "label": "Prerender (write)"},
        {"value": "CreateCamera", "label": "Camera (3d)"},
        {"value": "CreateGizmo", "label": "Gizmo (group)"},
        {"value": "CreateWriteImage", "label": "Image (write)"},
        {"value": "CreateModel", "label": "Model (3d)"},
        {"value": "CreateBackdrop", "label": "Nukenodes (backdrop)"},
        {"value": "CreateWriteRender", "label": "Render (write)"},
        {"value": "CreateSource", "label": "Source (read)"},
    ]

class NodesModel(BaseSettingsModel):
    _layout = "expanded"
    plugins: list[str] = SettingsField(
        default_factory=list,
        title="Used in plugins",
        enum_resolver=nuke_creator_plugins_enum
    )
    nuke_node_class: str = SettingsField(
        title="Nuke Node Class",
    )
```

</details>

Image

### Describe alternatives you've considered:

_No response_

MustafaJafar:
## Changelog Description
resolve #42 
Reuse the names/labels that admins/nuke artists are familiar with.
where the setting shows a drop down menu and select the creators

| Publisher | Settings |
|--|--|
| Image |  Image |

Should we do the same with `Nuke Node class`
image

## Additional Info 
I was modifying the code to take a neat screenshot but it turns out that the small modification actually works as before.

## Testing notes:
1. Creating Nuke node should work as before.

### DIFF
--- a/server/settings/imageio.py
+++ b/server/settings/imageio.py
@@ -11,12 +11,24 @@ from .common import (
     ColorspaceConfigurationModel,
 )
 
+def nuke_creator_plugins_enum():
+    return [
+        {"value": "CreateWritePrerender", "label": "Prerender (write)"},
+        {"value": "CreateCamera", "label": "Camera (3d)"},
+        {"value": "CreateGizmo", "label": "Gizmo (group)"},
+        {"value": "CreateWriteImage", "label": "Image (write)"},
+        {"value": "CreateModel", "label": "Model (3d)"},
+        {"value": "CreateBackdrop", "label": "Nukenodes (backdrop)"},
+        {"value": "CreateWriteRender", "label": "Render (write)"},
+        {"value": "CreateSource", "label": "Source (read)"},
+    ]
 
 class NodesModel(BaseSettingsModel):
     _layout = "expanded"
     plugins: list[str] = SettingsField(
         default_factory=list,
-        title="Used in plugins"
+        title="Used in plugins",
+        enum_resolver=nuke_creator_plugins_enum
     )
     nuke_node_class: str = SettingsField(
         title="Nuke Node Class",


MustafaJafar:
Also, I'd like to point out that when adding new creator plugins, we should update the two enums accordingly.

server/settings/imageio.py
```diff
@@ -11,15 +11,41 @@
...
+def nuke_node_class_enum():
+    return [
+        {"value": "Write", "label": "Write [Image]"},
+        {"value": "Read", "label": "Read [Image]"},
+        {"value": "Group", "label": "Group [Other]"},
+        {"value": "Camera4", "label": "Camera [3D]"},
+        {"value": "Camera2", "label": "Camera [3D Classic]"},
+        {"value": "Scene", "label": "Scene [3D Classic]"},
+        {"value": "BackdropNode", "label": "Backdrop [Other]"}, 
+    ]
```
robin-ynput, line 36: Potentially not for this PR, but I'm wondering if there is a way to make these enums dynamically retrieved ?
We would need to add a class attribute to each creator to get the associated Nuke node Class and gather the creators.

Benefit imo would be that we get server settings follow code adjustments.

robin-ynput (approved):
Confirmed that it worked and settings were properly reported.

image

PR #52, opened 2025-01-14
Fix use of stagingDir

kalisp:
## Changelog Description
Use stagingDir on instance only if it is explicitly set as custom, in other case fall back to old implementation which produces baking script next in `work/../renders`.

This solves issue in `extract_review_intermediates` when now there was `stagingDir` on instance suddenly. This resulted in creation of baking script in `tempdir` which got purged at the end of publishing via `Clean Up` plugin.

This PR is variant of similar issue in Maya https://github.com/ynput/ayon-maya/pull/212

## Testing notes:
1. publish in Nuke via Deadline
2. baking step shouldnt fail

### DIFF
--- a/client/ayon_nuke/plugins/publish/collect_nuke_instance_data.py
+++ b/client/ayon_nuke/plugins/publish/collect_nuke_instance_data.py
@@ -58,10 +58,14 @@ class CollectInstanceData(pyblish.api.InstancePlugin):
         staging_dir_persistent = instance.data["transientData"].get(
             "stagingDir_persistent", False
         )
+        staging_dir_is_custom = instance.data["transientData"].get(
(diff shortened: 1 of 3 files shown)


jakubjezek001 (approved):
Looking good, but I let @robin-ynput judge if this is a solid ground.

robin-ynput (approved):
@kalisp I hope you don't mind me pushing on the top of you PR.

When testing these changes, I realized Nuke staging directory was not working as expected after https://github.com/ynput/ayon-core/pull/1069, so I've added the necessary code changes.

Please note that it will also require this fix in `ayon-core` to behave properly:  https://github.com/ynput/ayon-core/pull/1084

kalisp:
Tested, works.

PR #58, opened 2025-01-21
Replace render path in write nodes from stagingDir

### LINKED ISSUE #55: AY-7342_Nuke: write nodes path doesn't recalculate

Path where Write node will render stays same after creation. This would cause an issue if it should be filled with custom staging dir template that contains {version} , eg if it should be rendering into versioned location.

This affects both local and farm renders.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1329049224804962304/1329432926865330269)

_This issue was automatically created from Clickup ticket [AY-7342](https://app.clickup.com/t/86c1nhw4v)_

kalisp:
## Changelog Description
Custom `stagingDir` (`ayon+settings://core/tools/publish/custom_staging_dir_profiles`) could use dynamic portion, using {version} or some unique descriptor. Render path of write nodes sets this `stagingDir` path only during creation time (eg. it stays static), this updates each publish run.

New publish plugin was created in `Other` section which updates version in render path to matching work workfile version (which got just bumped up step before). That way `renders` subfolders would be unique.

This plugin is optional and by default its disabled.
image

Use case for this would be to sending multiple jobs do Deadline with suspended Publish job. Artist then will visually check all rendered images and enable only appropriate `Publish` job.

~~Issue is how to build unique portion of path, `{version}` doesn't work as it points to published version of product. Time component  (like `{ddmmyyyy}`) doesn't seem to work.

Another issue is if path in write nodes should be refreshed all the time, or it should be configurable in Settings (and by default disabled). (Problem there is that one needs to remember prepare custom staging dir template AND enable it in `CollectNukeWrites`)~~

Resolved by using new integrator which rewrites writer node render path with use of collected version from workfile. It is applied only on render nodes which are set up with custom staging dir)

Should be tested together with https://github.com/ynput/ayon-deadline/pull/120

## Testing notes:
1. use custom staging dir template with dynamic part, eg {@version}
2. enable plugin in `ayon+settings://nuke/publish/IncrementWriteNodePath`
3. try couple of publishes, local and farm and check if path in write nodes changes

resolves #55

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -22,7 +22,6 @@ from ayon_core.pipeline.workfile.workfile_template_builder import (
 from ayon_core.lib import (
     env_value_to_bool,
     Logger,
-    get_version_from_path,
     StringTemplate,
 )
 
@@ -836,9 +835,12 @@ def check_inventory_versions():
         log.warning(error)
 
 
-def writes_version_sync():
-    ''' Callback synchronizing version of publishable write nodes
-    '''
+def writes_version_sync(write_node, log):
+    """ Callback synchronizing version of publishable write nodes
+
+    Tries to find version string in render path of write node and bump it to
+    workfile version.
+    """
     try:
         rootVersion = get_version_from_path(nuke.root().name())
         padding = len(rootVersion)
@@ -846,32 +848,50 @@ def writes_version_sync():
             int(rootVersion)
         )
     except Exception:
+        log.warning("exception", exc_info=True)
         return
 
-    for each in nuke.allNodes(filter="Write"):
-        # check if the node is avalon tracked
-        if NODE_TAB_NAME not in each.knobs():
-            continue
+    try:
+        write_path = write_node["file"].value()
+        node_version = "v" + get_version_from_path(write_path)
 
-        avalon_knob_data = read_avalon_data(each)
+        log.debug(f"Replacing '{node_version}' with '{new_version}'")
+        node_new_file = write_path.replace(node_version, new_version)
+        write_node["file"].setValue(node_new_file)
+        if not os.path.isdir(os.path.dirname(node_new_file)):
+            log.warning("Path does not exist! I am creating it.")
+            os.makedirs(os.path.dirname(node_new_file))
+    except Exception as e:
+        log.warning(
+            f"Write node: `{write_node.name()}` has no version "
+            f"in path: '{write_path}'. Expected format as `.vXXX` or `_vXXX`.",
+            exc_info=True
+        )
 
-        try:
-            if avalon_knob_data["families"] not in ["render"]:
-                continue
+def get_version_from_path(file):
+    """Find version number in file path string.
 
-            node_file = each["file"].value()
+    Looks for formats:
+    - `_v0001`
+    - `.v001`
+    - `/v001/` - difference from ayon-core.path_tools.get_version_from_path
 
-            node_version = "v" + get_version_from_path(node_file)
+    Args:
+        file (str): file path
 
-            node_new_file = node_file.replace(node_version, new_version)
-            each["file"].setValue(node_new_file)
-            if not os.path.isdir(os.path.dirname(node_new_file)):
-                log.warning("Path does not exist! I am creating it.")
-                os.makedirs(os.path.dirname(node_new_file))
-        except Exception as e:
-            log.warning(
-                "Write node: `{}` has no version in path: {}".format(
-                    each.name(), e))
+    Returns:
+        str: version number in string ('001')
+    """
+
+    pattern = re.compile(r"[\._/]v([0-9]+)", re.IGNORECASE)
+    try:
+        return pattern.findall(file)[-1]
+    except IndexError:
+        log.error(
+            "templates:get_version_from_workfile:"
+            "`{}` missing version string."
+            "Example `v004`".format(file)
+        )
 
 
 def version_up_script():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -864,9 +864,10 @@ def writes_version_sync(write_node, log):
         node_new_file = re.sub(r'%0*(\d+)d', replace_match, node_new_file)
 
         write_node["file"].setValue(node_new_file)
-        if not os.path.isdir(os.path.dirname(node_new_file)):
-            log.warning("Path does not exist! I am creating it.")
-            os.makedirs(os.path.dirname(node_new_file))
+        render_dir = os.path.dirname(node_new_file)
+        if not os.path.isdir(render_dir):
+            log.warning(f"Path '{render_dir}' does not exist! Creating it.")
+            os.makedirs(render_dir)
     except Exception:
         log.warning(
             f"Write node: `{write_node.name()}` has no version "
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -840,6 +840,11 @@ def writes_version_sync(write_node, log):
 
     Tries to find version string in render path of write node and bump it to
     workfile version.
+
+    Args:
+        write_node (nuke.Node)
+        log (logging.Logger) - logger to output messages into Publisher
+
     """
     try:
         rootVersion = get_version_from_path(nuke.root().name())
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -30,7 +30,10 @@ from ayon_core.settings import (
     get_current_project_settings,
 )
 from ayon_core.addon import AddonsManager
-from ayon_core.pipeline.template_data import get_template_data_with_names
+from ayon_core.pipeline.template_data import (
+    get_template_data_with_names,
+    get_template_data,
+)
 from ayon_core.pipeline import (
     Anatomy,
     registered_host,
@@ -40,6 +43,7 @@ from ayon_core.pipeline import (
     get_current_task_name,
     AYON_INSTANCE_ID,
     AVALON_INSTANCE_ID,
+    get_current_context,
 )
 from ayon_core.pipeline.load import filter_containers
 from ayon_core.pipeline.context_tools import (
@@ -49,7 +53,9 @@ from ayon_core.pipeline.colorspace import (
     get_current_context_imageio_config_preset
 )
 from ayon_core.pipeline.workfile import BuildWorkfile
-from . import gizmo_menu
+from ayon_core.resources import get_ayon_icon_filepath
+
+from .gizmo_menu import GizmoMenu
 from .constants import (
     ASSIST,
     LOADER_CATEGORY_COLORS,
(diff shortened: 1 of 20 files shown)


BigRoy:
> Issue is how to build unique portion of path, `{version}` doesn't work as it points to published version of product.

`{workfile_version}` could work though - although it'd require the workfile to ALWAYS have an actual version which should be 99% f the time though? Also, I really like just using the workfile name as "unique" token so a `{workfile}` (which would be without extension) or similar could also solve this

kalisp:
Question: Does limiting path rewrites for `frames*` workflows make sense? (Eg. if artist select "Use existing frame" it shouldnt replace render path in a write node, as workfile, eg. version, would get bumped automatically.)
If thats the case, `writes_version_sync` must contain additional argument to apply on specific write node.

I dont fully understand `frames` workflow, so not sure about all implications.

kalisp:
`writes_version_sync` needs to be translated into AYON format, it is obsolete.

Move logic to extractor phase after workfile version is bumped up.

server/settings/publish_plugins.py
```diff
@@ -285,10 +279,25 @@ class PublishPluginsModel(BaseSettingsModel):
...
     )
-    IncrementScriptVersion: IncrementScriptVersionModel = SettingsField(
+    IncrementScriptVersion: OptionalPluginModel = SettingsField(
         title="Increment Workfile Version",
-        default_factory=IncrementScriptVersionModel,
-        section="Integrators"
+        default_factory=OptionalPluginModel,
+        section="Integrators",
+        description=(
+            "Bumps up version of workfile if all no errors in previous "
```
BigRoy, line 287:
```suggestion
            "Bumps up version of workfile if there are no errors in previous "
```

jakubjezek001 (approved):
This way it works perfectly!

kalisp:
Baking script needs to be checked.

kalisp:
Checkout https://github.com/ynput/ayon-deadline/pull/120 to fully test it. (DL publish will fail without it.)

jakubjezek001 (changes requested):
The path in write node was incremented (synced with workfile) even I had disabled it with the new plugin attribute. 

image

kalisp:
> The path in write node was incremented (synced with workfile) even I had disabled it with the new plugin attribute.
> 
Fixed.

client/ayon_nuke/api/plugin.py
```diff
@@ -783,6 +789,32 @@ def get_imageio_baking_profile(self):
...
 
+    def _update_staging_dir(self, current_file, staging_dir):
+        """Update staging dir with current file version.
+
+        Staging dir is used as a place where intermediate review files should
+        be stored. If render path contains version portion, which is replaced
+        by version from workfile, it must be reflected even for baking scripts.
+        """
+        try:
+            rootVersion = get_version_from_path(current_file)
```
robin-ynput, line 800: I just want to flag a small caveat I encounter with this approach (always replace by version from workfile).
Here is what I've done
* disable Follow Workfile Version `ayon+settings://core/publish/CollectAnatomyInstanceData/follow_workfile_version` 
* setup a staging directory template
* create a new shot and bump the workfile to `v3` but no publish
* try to publish a new render (that should become v1) producing a `review`.

The `review` part failed with the following traceback:
```
DEBUG: Path: C:/projects/Dummy/Nuke/test_staging/version_staging/temp_render_only/render_new/renderCompositingMain_v004/renderCompositingMain.baking.mov
DEBUG: Write...   `[<Read2 at 0x000001A4A9C5E7F0>, <OCIOColorSpace2 at 0x000001A4A9C5EAF0>, <Write2 at 0x000001A4A9C5EB50>]`
INFO: Rendering...  
  File "C:\Users\robin\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-nuke\client\ayon_nuke\plugins\publish\extract_review_intermediates.py", line 139, in process
    data = exporter.generate_mov(delete=delete, **o_data)
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-nuke\client\ayon_nuke\api\plugin.py", line 1210, in generate_mov
    self.render(write_node.name())
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-nuke\client\ayon_nuke\api\plugin.py", line 999, in render
    nuke.execute(
RuntimeError: Write2: unable to open file
```
I believe this is because `extract_review_intermediates` use this new logic to refresh its staging directory based on the workfile version (`v3`). 
But in my case this directory did not exist as `follow_workfile_version` was disabled and it was expected to create `v1`. 

BTW this seems to be only a problem here because when `review` is disabled, the render itself publishes correctly under `v1`

Here is a more detailed publish error for my second attempt (v4 vs v2): 
[publish-report-250211-12-39.json](https://github.com/user-attachments/files/18749327/publish-report-250211-12-39.json)

client/ayon_nuke/api/plugin.py
```diff
@@ -783,6 +789,32 @@ def get_imageio_baking_profile(self):
...
+                f"to {new_version}"
+            )
+            return staging_dir.replace(staging_dir_version, new_version)
+        except IndexError:
+            self.log.warning(
+                f"Current file '{current_file}' or staging_dir "
+                f"'{staging_dir}' don't contain version number. "
+                "No replacement necessary",
+                exc_info=True)
+            return
```
robin-ynput, line 817: I think it shouldn't return None if nothing needs to be done but untouched `staging_dir`.
Also I tried with a malformed template `foo.{version}` instead of `foo.v{version}` and it failed with a `TypeError`.

```suggestion
    try:
        root_version = get_version_from_path(current_file)
        padding = len(root_version )
        root_version = int(root_version)
    except (TypeError, IndexError):
        self.log.warning(
            f"Current file '{current_file}' doesn't contain version number. "
            "No replacement necessary",
            exc_info=True)
        return staging_dir

    try:
        staging_dir_version = "v" + get_version_from_path(staging_dir)
    except (TypeError, IndexError):
        self.log.warning(
            f"Staging directory '{staging_dir}' doesn't contain version number. "
            "No replacement necessary",
            exc_info=True)
        return staging_dir   
                     
    new_version = "v" + str("{" + ":0>{}".format(padding) + "}").format(
                rootVersion
            )
    self.log.debug(
        f"Update version in staging dir from {staging_dir_version} "
        f"to {new_version}"
    )
    return staging_dir.replace(staging_dir_version, new_version)

```

robin-ynput (changes requested):

client/ayon_nuke/api/plugin.py
kalisp: Ok, makes sense.

BigRoy:
Not sure if Nuke supports it - but this would become way better if in some way we could support an environment variable in the filepath, like e.g. `$AYON_WORKFILE_VERSION`. Then all we'd need to do is to make sure we have the env var set accordingly and correctly all the time - which may make this much more 'generic' and potentially applicable to more host applications too.

But might be just me. I'm really not a fan of just `str.replace("v001", "v002")` because `dev001_v001` would suddenly become `dev002_v002` to. The chance to side effects just seem a bit dangerous.

By using an _actual_ variable inside output paths it means we don't actually need to be adjusting filepaths at all over time, it'd be dynamic by its nature.

client/ayon_nuke/api/plugin.py
kalisp: Couldnt reproduce, problem is nonexistence of path in Write node (tried that, deleted that version `renders` subfolder and got recreated).
I dont see why it would fail this way on this place.

client/ayon_nuke/api/lib.py
```diff
@@ -836,42 +835,70 @@ def check_inventory_versions():
...
 
 
-def writes_version_sync():
-    ''' Callback synchronizing version of publishable write nodes
-    '''
+def writes_version_sync(write_node, log):
+    """ Callback synchronizing version of publishable write nodes
+
+    Tries to find version string in render path of write node and bump it to
+    workfile version.
```
jakubjezek001, line 842: we should add Attributes please

client/ayon_nuke/api/plugin.py
jakubjezek001: well from the logic of this implementation this should not be happening. The `writes_version_sync` function is really just synchronizing to current workfile version. This function is triggered post integration plugin. But your case @robin-ynput should be definitely replicable if it is happening to you. It seems that the expected path from the **ExtractReviewIntermediates** is different from what is actually set in the current write render node and we are resolving it dynamically. @kalisp this might be possible since the path is inherited form `instance.data["path"]` look at line 45. I am not currently sure where is the path key collected from, but it might be different path then the write node collected.

tweak-wtf:
i have some time now to test this aswell.

> It'd be way more explicit if the actual write path contains e.g. `${version}` or whatever form of 'expressions' or variables it may support.

yes this is actually possible by using inline TCL expression in the write node's `file` knob. i just checked
so something like this here is totally doable and resolves.
`C:/Users/artist/cool/render/folder/[getenv AYON_WORKFILE_VERSION]/file.####.exr`
note the `[getenv AYON_WORKFILE_VERSION]`

client/ayon_nuke/api/plugin.py
tweak-wtf: i think i can somewhat reproduce this.
it's a bit odd to test `Follow Workfile Version` and then disable it in the first place but i see the usecase for this wanting to resolve `{@version}` with disbaled `follow workfile version`.
my last published render product was at v025, so workfile at v026. i incremented twice to workfile v028 and submitted my render to deadline expecting v026 to be published.

the render/publish went through fine for me, but the baking job rendered to a v028 folder. so it was following workfile version although disabled. basically baking/reviews were rendered to v028 folder but integrated/published to v026 as expected.

not sure whether the baking nuke job should live in a v026 folder in this case. right now, the baking nkscript live next to the original one in a v028 folder.

tweak-wtf:
testing this i noticed some details:
running `IncrementWriteNodePath` at the very end causes jobs that are rendered on the farm with `Use Published Workfile` to not pick up those changed output paths in the published workfile and in turn renders to an older/unexpected folder. running it at default order of `pyblish.api.IntegratorOrder` fixed it for me.

after successful publish when the workfile gets incremented the output paths should also be updated accordingly. currently they don't get updated which causes write nodes to render to the previously used folder which can cause confusion. This only happens if a user would use the local render button from the AYONWriteNode directly. Using the publisher increments path tokens correctly.

during cleanup phase the baking nkscript doesn't get removed. i believe the parent folder should also get deleted as at this point the baking nkscript is the only file in there

kalisp:
> testing this i noticed some details: running `IncrementWriteNodePath` at the very end causes jobs that are rendered on the farm with `Use Published Workfile` to not pick up those changed output paths in the published workfile and in turn renders to an older/unexpected folder. running it at default order of `pyblish.api.IntegratorOrder` fixed it for me.

Just double-checking if you are testing this with https://github.com/ynput/ayon-deadline/pull/120 . Not sure if it applies, but `Use Published Workfile` sounds weirdly familiar.

tweak-wtf:
> Just double-checking if you are testing this with [ynput/ayon-deadline#120](https://github.com/ynput/ayon-deadline/pull/120) . Not sure if it applies, but `Use Published Workfile` sounds weirdly familiar.

just double-checked this on my end but yeah. it does happen with ayon-deadline at 0.5.4.
because the write node paths get updated after the workfile is published

kalisp:
I am not sure I follow, it should paths with `work - workfile` version, so even if `Use Published` is ticked, it will not use version from `publish - workfile`. It should? 

If that's the case, your solution will not help as it will still uses `work - workfile` version, albeit incremented. There is high probability, that `publish - workfile` version will not have same value, most likely `work - workfile` will be couple of versions ahead as some errors could happen after workfile version increment, this happens to me all the time.

To solve this we would need to parse out version from `publish - workfile` if `Use Published` is used.

BigRoy:
> it will not use version from `publish - workfile`. It should?

It should use the _workfile version_ I'd say.

tweak-wtf:
> it should paths with work - workfile version, so even if Use Published is ticked, it will not use version from publish - workfile. It should?

jep i think so. thing is that afaik on deadline nodes there is no logic happening in nuke, besides GlobalJobPreLoad. it just gets a nkscript and a write node, off it goes.
so if the write node's paths don't get updated before the "work nkscript" is published and integrated (which is used if `Use Published Scene` is enabled), they contain the old version.

> Assuming that the "new version" is the new intermediate version for next time - not the version number you want to render now.

yes exactly. that "new version" is basically the version of the soon to be incremented workfile, or product based on write node if `Follow Workfile Version` is disabled.
that's why i wanted to execute the `writes_version_sync` a second time in the end. to make sure a user's workfile is prepped after it got incremented. just a little qol but all good, i can just add it to our fork or studio addon.

philippe-ynput:
@kalisp Can you please wrap this up when you have a moment ?

kalisp:
I updated it with Tony's code. Tested it, seemed fine.

jakubjezek001 (approved):
looking good

tweak-wtf (approved):
that does its job 👌

PR #61, opened 2025-01-27
Fix retime gathering from Nuke (freeze frame, linear speed and TimeWarp)

robin-ynput:
## Changelog Description

Resolve: https://github.com/ynput/ayon-hiero/issues/12
Resolve: https://github.com/ynput/ayon-hiero/issues/29

Adjusted:
* [X] Fix linear speed gathering and freeze frames
* [X] Ensure timewarps are properly retrieved

## Additional notes
These changes needs to be tested alongside https://github.com/ynput/ayon-core/pull/1109

## Testing notes:
1. Publish plates with `speed-change (positive+negative)`, `freeze frames` and `TimeWarp` from an Editorial DCC
2. Gather published plates in `Nuke`
3. Ensure the working ranges match

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -240,7 +240,7 @@ class LoadClip(plugin.NukeLoader):
                 data=data_imprint)
 
         if add_retime and version_data.get("retime"):
-            self._make_retimes(read_node, version_data)
+            self._make_retimes(read_node, version_attributes, version_data)
 
         self.set_as_member(read_node)
 
@@ -387,7 +387,7 @@ class LoadClip(plugin.NukeLoader):
             )
 
         if add_retime and version_data.get("retime"):
-            self._make_retimes(read_node, version_data)
+            self._make_retimes(read_node, version_attributes, version_data)
         else:
             self.clear_members(read_node)
 
@@ -451,7 +451,7 @@ class LoadClip(plugin.NukeLoader):
 
             read_node['frame'].setValue(str(start_frame))
 
-    def _make_retimes(self, parent_node, version_data):
+    def _make_retimes(self, parent_node, version_attributes, version_data):
         ''' Create all retime and timewarping nodes with copied animation '''
         speed = version_data.get('speed', 1)
         time_warp_nodes = version_data.get('timewarps', [])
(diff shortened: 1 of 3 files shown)


server/settings/loader_plugins.py
```diff
@@ -66,7 +66,7 @@ class LoaderPluginsModel(BaseSettingsModel):
         "representations_include": [],
         "node_name_template": "{class_name}_{ext}",
         "options_defaults": {
-            "start_at_workfile": True,
+            "start_at_workfile": False,
```
robin-ynput, line 69: @jakubjezek001 I don't understand the logic behind this option, it's coming initially from OpenPype. 

From what I could see it is only used in `Clip` loader (not `Image`) and forces the created `Read` node to set `frame_mode -> startat -> workfile first frame`.

With this enabled, it creates an offset so the first handle frame (e.g. 981) becomes the workfile start frame (e.g. 1001).
https://learn.foundry.com/nuke/content/reference_guide/image_nodes/read.html

I had to turn it off for my testing because this offset resulted in:
- the working range in Nuke does not match the extracted plate range anymore.
- all applied `Retime` and `TimeWarp` nodes, which are based on the extracted plate range, becomes wrong

 https://learn.foundry.com/nuke/content/reference_guide/image_nodes/read.html

jakubjezek001: this option was added for the described reason. But usually all image sequences are renumbered to fit first workfile frame number so perhaps this is really not needed to be set active by default.

jakubjezek001: btw, I was solving the issue other way around. The Read had been set to Start at 1001 and then:
- retime input range start toggle deactivated
- retime input range end toggle deactivated

This had offseted the clip range to correct frames. 

image

jakubjezek001 (approved):
works perfectly

client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -469,14 +469,22 @@ def _make_retimes(self, parent_node, version_data):
...
+                )
 
                 rtn["before"].setValue("continue")
                 rtn["after"].setValue("continue")
+                rtn["reverse"].setValue(speed < 0)
+
                 rtn["input.first_lock"].setValue(True)
                 rtn["input.first"].setValue(
-                    self.script_start
+                    version_attributes["frameStart"]
```
iLLiCiTiT, line 481: Is it ok that the attribute `"frameStart"` on version does not have to be filled during publishing?

In that case the value constains whatever is set on it's parent folder.

server/settings/loader_plugins.py
robin-ynput: I'll follow your advise and keep it disabled by default.

client/ayon_nuke/plugins/load/load_clip.py
robin-ynput: Correct me if I'm wrong, but I believe for `clip` products those values are part of the publishing.
* `frameStart` and `frameEnd` are always set to `instance.data`  [here](https://github.com/ynput/ayon-core/blob/d90324313cdd94af0a4fbd55d56026ca1f805290/client/ayon_core/plugins/publish/collect_otio_frame_ranges.py#L78)
* `frameStart` and `frameEnd` are then set from `instance.data` to `version_attributes` [here](https://github.com/ynput/ayon-core/blob/d90324313cdd94af0a4fbd55d56026ca1f805290/client/ayon_core/plugins/publish/integrate.py#L454)

So for `clip` produts we should be covered, aren't we ?

iLLiCiTiT: > Correct me if I'm wrong, but I believe for clip products those values are part of the publishing.

I see potential danger 🙂 , and possibility of using wrong data. But if it's covered, then it is ok.

robin-ynput:
Waiting for official release of https://github.com/ynput/ayon-core/pull/1087 to update requirements and merge.

PR #64, opened 2025-01-31
Update colorspace retrieval logic

jakubjezek001:
## Changelog Description
- Enhanced colorspace fetching to prioritize new data structure.
- Added fallback options for backward compatibility.
- Improved logging for better debugging of colorspace sources.## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

Loading Image was still in old deprecated way of colorspace maintaining.

## Testing notes:
1. load with clip or image loader
2. notice that the colorspace set to node is the one defined in representations' colorspaceData

### DIFF
--- a/client/ayon_nuke/plugins/load/load_image.py
+++ b/client/ayon_nuke/plugins/load/load_image.py
@@ -10,10 +10,15 @@ from ayon_core.pipeline import (
 from ayon_nuke.api.lib import (
     get_imageio_input_colorspace
 )
+from ayon_core.pipeline.colorspace import (
+    get_imageio_file_rules_colorspace_from_filepath,
+    get_current_context_imageio_config_preset,
+)
 from ayon_nuke.api import (
     containerise,
     update_container,
-    viewer_update_and_undo_stop
+    viewer_update_and_undo_stop,
+    colorspace_exists_on_node,
 )
 from ayon_core.lib.transcoding import (
     IMAGE_EXTENSIONS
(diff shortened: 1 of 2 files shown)


client/ayon_nuke/plugins/load/load_image.py
```diff
@@ -156,33 +160,33 @@ def update(self, container, context):
...
+        assert read_node.Class() == "Read", "Must be Read"
 
         project_name = context["project"]["name"]
         version_entity = context["version"]
         repre_entity = context["representation"]
 
         repr_cont = repre_entity["context"]
 
-        file = get_representation_path(repre_entity)
+        filepath = get_representation_path(repre_entity)
```
moonyuet, line 174: not related to this PR but we can replace this function as suggested below, if you want to support the loaders across different project. 
```suggestion
        filepath = self.filepath_from_context(context)
```

moonyuet (changes requested):
I tested with updating and switching function with the clip loaded by the old clip loader. Works nicely.
I also tested with loading clip/image function, loading clip works as expected to load the clip with correct colorspace data. But the image loader still gives colorspace error for loading single image frame.
image

moonyuet:
> > But the image loader still gives colorspace error for loading single image frame.
> 
> @moonyuet I tried to load a single image frame (clip + image) from single image plate and I didn't get any issue. Do you have any error message or any particular context to generate this single image frame ?

No not really. I believe it could be because I am using the old scene to do so.
I re-tested it again with loading image with the texture and also the same clips. they are all working now. 
image

PR #65, opened 2025-01-31
Improve selection handling in NukeWriteCreator

### LINKED ISSUE #56: AY-7420_Templated Workfile Builder doesn't create Write

## Current Behavior
With update to Core 1.0.13 and Nuke 0.3.0, Nuke placeholder for create_write_render creator doesn't work. The same placeholder works as expected (makes Ayon write group) with  Core 1.0.12 and Nuke 0.2.6 addons.

Image

### Steps To Reproduce:

1. Make template with placeholder to create write render
2. Apply the template. No error message, and no write group created

### Version
Server 1.6.0, Core 1.0.13 and Nuke 0.3.0

Windows 10
Nuke 14.0v4

### Relevant log output:

```shell

```

jakubjezek001:
## Changelog Description
- Added early return for no selection case.
- Simplified logic to raise error only if selected nodes are missing.
- Handled headless mode by defaulting pre_create_data when absent.

closes #56 

## Testing steps
1. have activated workfile template building templates at your project
2. have Create placeholder in template
3. build templates and see that creator placeholder is working again.

### DIFF
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -360,12 +360,15 @@ class NukeWriteCreator(NukeCreator):
         Raises:
             NukeCreatorError. When the selection contains more than 1 Write node.
         """
+        if not pre_create_data.get("use_selection"):
+            return []
+
         selected_nodes = super()._get_current_selected_nodes(
             pre_create_data,
             class_name=None,
         )
 
-        if pre_create_data.get("use_selection") and not selected_nodes:
+        if not selected_nodes:
             raise NukeCreatorError("No active selection")
 
         elif len(selected_nodes) > 1:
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/api/plugin.py
```diff
@@ -452,6 +455,12 @@ def _get_render_target_enum(self):
         )
 
     def create(self, product_name, instance_data, pre_create_data):
+        if not pre_create_data:
+            # add no selection for headless
+            pre_create_data = {
+                "use_selection": False
+            }
```
BigRoy, line 462: This seems odd to do. Especially because this doesn't necessarily differentiate between "headless" (e.g. Nuke without UI) or running the creator through a script, like e.g. template placeholders.

If the issue is to solve template placeholders, then you may be better off making sure that the Loader used there actually clears the selection?

robin-ynput:
The code looks OK and confirmed I have been able to test it locally.

robin-ynput (approved):

PR #77, opened 2025-02-19
Make ocioLook loader handle multiple chained LUTs.

robin-ynput:
## Changelog Description

Changes on the Nuke `ociolook` loader:
* Ensure multiple chained LUT published are properly loaded
* Fix the `invert_direction` LUT attribute for the `OCIOFileTransform` nuke node

This changes would be backward compatible with `ociolook` product from previous iterations.
To test the multiple/chained LUT feature, please refer to : https://github.com/ynput/ayon-traypublisher/pull/71

## Testing notes:
1. Publish some `ociolook` product via the TrayPublisher (https://github.com/ynput/ayon-traypublisher/pull/71)
2. Load the resulting product in Nuke through the loader

### DIFF
--- a/client/ayon_nuke/plugins/load/load_ociolook.py
+++ b/client/ayon_nuke/plugins/load/load_ociolook.py
@@ -180,7 +180,7 @@ class LoadOcioLookNodes(load.LoaderPlugin):
                     (
                         file for file in all_files
                         if file.endswith(extension)
-                        and lut_suffix in os.path.basename(file)
+                        and os.path.basename(file).endswith(lut_suffix)
                     ),
                     None
                 )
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/plugins/load/load_ociolook.py
```diff
@@ -174,11 +174,13 @@ def _create_group_node(
                 # file path from lut representation
                 extension = ocio_item["ext"]
                 item_name = ocio_item["name"]
+                lut_suffix = ocio_item.get("lut_suffix", "")
 
                 item_lut_file = next(
                     (
                         file for file in all_files
                         if file.endswith(extension)
+                        and lut_suffix in os.path.basename(file)
```
iLLiCiTiT, line 183: Shouldn't that be explicit suffix check?

I guess we do modify the file so the suffix might not be at the end of filename, or?
```suggestion
                        and os.path.basename(file).endswith(lut_suffix)
```

jakubjezek001: Ok so I have forgotten that the suffix sill be added to the lut file during integration since the `outputName` is added to the representation. In this case this is actually working . So yes this is working and your suggestion should too @iLLiCiTiT .

jakubjezek001 (approved):
works well

iLLiCiTiT (approved):

Issue #74, opened 2025-02-12
AY-7475_Set default render target from local to farm


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

we should have possibility to define default render target

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1339180322478362646/1339186365493018645)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7475](https://app.clickup.com/t/86c22mftr)_

PR #80, opened 2025-02-25
AY-7475 - Define default render targets through settings.

robin-ynput:
## Changelog Description

Resolve: #74 
Add new settings to be able to define default render targets for `Render`, `Image` and `Prerender` creators

image

## Testing notes:
1. Create new package and upload it to AYON server
2. Edit newly created settings:
    * `ayon+settings://nuke/create/CreateWriteImage/render_target`
    * `ayon+settings://nuke/create/CreateWritePrerender/render_target`
    * `ayon+settings://nuke/create/CreateWriteRender/render_target`

3. In Nuke open `Creator` widget and ensure default render target(s) are properly gathered from settings


client/ayon_nuke/api/plugin.py
```diff
@@ -438,16 +440,24 @@ def _get_render_target_enum(self):
...
+        farm_rendering_targets = {
+            "frames_farm": "Use existing frames - farm",
+            "farm": "Farm rendering",
+        }
+
+        if (
+            "farm_rendering" in self.instance_attributes
+            or self.render_target in farm_rendering_targets
+        ):
+            rendering_targets.update(farm_rendering_targets)
```
iLLiCiTiT, line 453: I don't think we should add farm options just because you have farm value set in settings as default -> it's hiding wrong settings.
```suggestion
        if "farm_rendering" in self.instance_attributes:
            rendering_targets.update(farm_rendering_targets)
```

iLLiCiTiT, line 453: I don't think we should add farm options just because you have farm value set in settings as default -> it's hiding wrong settings.
```suggestion
        if "farm_rendering" in self.instance_attributes:
            rendering_targets.update({
                "frames_farm": "Use existing frames - farm",
                "farm": "Farm rendering",
            })
```

robin-ynput: On the other end, wouldn't it be confusing for the user to be able to set `Farm rendering` in the settings but not seing it applied and available because the `farm_rendering` attribute is missing ?

or I can make it raises with a `CreatorError` if both options conflicts.

client/ayon_nuke/api/plugin.py
```diff
@@ -438,16 +440,33 @@ def _get_render_target_enum(self):
...
+        farm_rendering_targets = {
+            "frames_farm": "Use existing frames - farm",
+            "farm": "Farm rendering",
+        }
+
+        if (
+            self.render_target in farm_rendering_targets
+            and "farm_rendering" not in self.instance_attributes
+        ):
+            raise NukeCreatorError(
```
iLLiCiTiT, line 453: This could be raised on collection of instances, user will never know about it.

BigRoy:
Please confirm @jakubjezek001 :)

robin-ynput:
Thanks guys for the push on this one 🚀

PR #93, opened 2025-03-11
Load GeoImport/GeoReference: Add support for loading USD or Alembic using `GeoImport` and `GeoReference`

BigRoy:
## Changelog Description

Add support for loading into Nuke's New 3D System USD-Based Nodes, specifically `GeoImport` and `GeoReference` nodes currently in Beta.

Should work fine for Nuke 16 - but is also available in 15.2. I'm not entirely sure from which version the new 3D system was implemented.

The `GeoImport` is essentially a USD Sublayer.
The `GeoReference` is essentially a USD Reference or Payload.

See:
- https://learn.foundry.com/nuke/content/reference_guide/3dnew_nodes/usd3d_nodes.html
- https://learn.foundry.com/nuke/content/reference_guide/3dnew_nodes/geoimport.html
- https://learn.foundry.com/nuke/16.0/content/reference_guide/3dnew_nodes/georeference.html

## Testing notes:

1. Loading USD and Alembic should wokr
2. Updating should work


BigRoy:
Seems to work for me.

philippe-ynput:
LGTM, but...
Can we get someone else to test this first ? @MustafaJafar maybe ?

MustafaJafar (approved):
This is cool. 
It works on my side Nuke `14.0v5`.

image

BigRoy:
@jakubjezek001 let me know if you like it this way - it may clutter the "load" options for those who may not expect this to be added and/or may not actively be using these in production.

I wonder if we should at least still tweak the `order` to make sure they are less or more prioritized?
And I wonder if we should make it possible to enable/disable these loaders in settings?

jakubjezek001:
> I wonder if we should at least still tweak the `order` to make sure they are less or more prioritized?

Yes please. It might be confusing for some. Lets shift this lower then "Load Alembic", just not to break someone's muscle memory.

> And I wonder if we should make it possible to enable/disable these loaders in settings?

That would also improve the workflow. I wonder whether we should do this on all plugins in separated PR. This PR should only contain settings for this plugin.

BigRoy:
> > I wonder if we should at least still tweak the `order` to make sure they are less or more prioritized?
> 
> Yes please. It might be confusing for some. Lets shift this lower then "Load Alembic", just not to break someone's muscle memory.
> 
> > And I wonder if we should make it possible to enable/disable these loaders in settings?
> 
> That would also improve the workflow. I wonder whether we should do this on all plugins in separated PR. This PR should only contain settings for this plugin.

Thanks, I re-ordered them.

So they come after the originally Load Geo:

image

And added settings:

image

So that when e.g. disabled the entry is not available:

image

They are enabled by default currently though.

jakubjezek001 (approved):
perfect, thanks!

PR #107, opened 2025-04-15
AY-7619 Fill gaps in extract review - Nuke

kalisp:
## Changelog Description
This PR contains Nuke specific handling of missing frames in review.

Only selected frames could be rendered in Deadline, Nuke review process needs to handle missing frames somehow. 

This affects Deadline renders only!

As Nuke has additional review producing scripts and not only generic `ExtractReview` one, this approach is necessary.

There is no easy way how to fill/limit missing frames fully as `ExtractReview` does. There is no way how to copy frames from previous version, nor it is possible to create review only from rendered files.

To fully implement same behavior would be quite an effort as baking process would need to be reworked from scratch.

This has relation to:
- https://github.com/ynput/ayon-core/pull/1234
- https://github.com/ynput/ayon-deadline/pull/139

Initial issue: [AY-7619_Publishing of incomplete sequence](https://github.com/ynput/ayon-core/issues/1205)

## Testing notes:
1. experiment with `ayon+settings://nuke/publish/ExtractReviewIntermediates/outputs/0/fill_missing_frames`
image
3. follow this step


64qam:
In Nuke, it works on farm. Not sure how it should behave when set `Fail on missing`. 
Would it be possible to have the same labels and order as in Core:
image
While Nuke has these:
image


Local render ignores Custom frames completely.

kalisp:
`Fail on missing` should be original implementation, where gaps were not filled at all. In that case Nuke publishing would fail (most likely).

I reordered options a bit to more match to ayon-core, but kept original behavior on top as for most users it will be this option selected.
image

Reason why both enums are not 1-to-1 is mentioned in Description. (But tldr: in Nuke we can only use internal Nuke options and that matches to enum in Nuke - no use previous version. In other DCC I cannot create `checkerboard`.)

This PR affects only Deadline rendering. Field for `Custom frames` wouldn't be shown at all for local rendering, as Nuke (and I think other DCC neither) can control it locally.
(But that would be part of much larger refactoring effort https://github.com/ynput/ayon-deadline/issues/57 )

64qam (approved):

MilaKudr (approved):
Tested in Nuke + Deadline, all options work properly.

PR #112, opened 2025-06-03
Fix exposed knobs on Write nodes.

rdelillo:
## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

Detailed information of the changes made to the product or service, providing an in-depth description of the updates and enhancements. This can include technical information, code examples and anything else needed for the review of the PR.

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -666,7 +666,7 @@ def get_node_path(path, padding=4):
 
 
 def get_nuke_imageio_settings():
-    return get_project_settings(Context.project_name)["nuke"]["imageio"]
+    return DirmapCache.project_settings()["nuke"]["imageio"]
 
 
 def get_imageio_node_setting(node_class, plugin_name, product_name):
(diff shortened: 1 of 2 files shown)


client/ayon_nuke/plugins/publish/validate_write_nodes.py
```diff
@@ -81,6 +87,13 @@ def process(self, instance):
         if write_node is None:
             return
 
+        # gather exposed knobs to remove them from knobs check.
+        nuke_settings = instance.context.data["project_settings"]["nuke"]
+        product_type = instance.data["productType"]
+        plugin = self.product_types_mapping[product_type]
+        create_settings = nuke_settings["create"][plugin]
+        exposed_knobs = create_settings.get("exposed_knobs", [])
```
BigRoy, line 95: Not sure how many knobs can be exposed - but you may want to turn this into a set if you run this _per knob_ so that the lookup is faster.

client/ayon_nuke/plugins/publish/validate_write_nodes.py
```diff
@@ -137,8 +150,8 @@ def process(self, instance):
 
             if (
                 node_value not in fixed_values
-                and key != "file"
-                and key != "tile_color"
+                and key not in exposed_knobs
+                and key not in ("file", "tile_color")
```
BigRoy, line 154: If we're skipping anyways - can't we just move this skip logic directly to line 123 and just `continue` as soon as we find anything we can skip. It saves us from having to try to get a values and doing conversions for things we're skipping anyway?

client/ayon_nuke/api/lib.py
```diff
@@ -666,7 +666,7 @@ def get_node_path(path, padding=4):
 
 
 def get_nuke_imageio_settings():
-    return get_project_settings(Context.project_name)["nuke"]["imageio"]
+    return DirmapCache.project_settings()["nuke"]["imageio"]
```
BigRoy, line 669: If we're having issues with this being slow - shouldn't just `get_project_settings` use a cache... and also, doesn't it actually itself already cache? :) Seems odd to go trough a `DirmapCache`?

rdelillo: I had another look and you're right `ayon_core.settings.get_project_settings` is already caching the information.
(I dont think the additional `DirmapCache` caching is relevant then, but I'll leave that to another PR).

client/ayon_nuke/plugins/publish/validate_write_nodes.py
rdelillo: Yeah good call, I have adjusted that properly.

jakubjezek001 (approved):

Issue #123, opened 2025-07-03
Wrong frame being integrated as slate frame


JoseCovaOS:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior

When publishing using `use existing frames - farm` option, during `IntegrateAsset` plugin execution, the slate node is not correctly being integrated.  It seems the collected writes list is wrong.

Here a piece of the deadline log:

```
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file ... /work/renderScanCheckMain.1001.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1000.exr
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file ... /work/renderScanCheckMain.1002.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1001.exr
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file... /work/renderScanCheckMain.1003.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1002.exr
...
```

## Expected Behavior

The copied file from `work` to `publish` should match frame numbering, in the example above, the log should be like this: 

``` 
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file ... /work/renderScanCheckMain.1000.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1000.exr
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file ... /work/renderScanCheckMain.1001.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1001.exr
STDOUT: DEBUG:pyblish.IntegrateAsset:Copying file... /work/renderScanCheckMain.1002.exr -> /publish/renderScanCheckMain/v002/renderScanCheckMain_v002.1002.exr
....
```

### Steps To Reproduce:

1.  Open Nuke.
2. Publish using `use existing frames - farm`
3. Check exrs in `publish` directory, the exr corresponding to the slate doesn't match the exr under `work`.

### Version
AYON server version: `1.9.2+202506031436`
AYON core version:  `1.3.2`
AYON nuke version: `0.3.5`

### What platform were you running when you found the bug?

Linux - Rocky8

### Relevant log output:

```shell

```

PR #124, opened 2025-07-03
Fix wrong frame being integrated as slate frame

JoseCovaOS:
[Here](https://github.com/ynput/ayon-nuke/issues/123) the related issue.

## Changelog Description

Issue comes from the following line: 

```
           slate_frame = collected_frames[0].replace(
                frame_start_str, frame_slate_str)
```

Since `collected_frames` was not sorted, the first element (`collected_frames[0]`) does not match the value of `frame_start_str` (which in our case corresponds to 1001). Therefore, replacing that value did not work and the following line `collected_frames.insert(0, slate_frame)` introduced whichever element is at `collected_frames[0]`. 

For instance, in our test, the first element was frame 1017 and `collected_frames` ended up being like this: 
```
COLLECTED FRAMES -> ['renderScanCheckMain.1017.exr', 'renderScanCheckMain.1017.exr', 'renderScanCheckMain.1051.exr',  ... ]
```

We now look for `frame_start_str` inside `collected_frames`, get its index and use it in the replace function.

Another possible fix would be just to sort `collected_frames` (change [this](https://github.com/ynput/ayon-nuke/blob/develop/client/ayon_nuke/plugins/publish/collect_writes.py#L408) line to `return sorted(collected_frames)` but I think it would be more consistent to look for the index and replace rather than assuming again index 0 is the frame start. 

## Testing notes:
1. Launch Nuke.
2. Generate exrs (with a slate frame)
3. Publish using `use existing frames - farm` option. 

Slate node should be integrated correctly.
closes https://github.com/ynput/ayon-nuke/issues/123

### DIFF
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -1,4 +1,5 @@
 import os
+import re
 import nuke
 
 import pyblish.api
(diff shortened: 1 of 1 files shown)


jakubjezek001:
closes https://github.com/ynput/ayon-nuke/issues/123

PR #130, opened 2025-07-21
Load: Handles on version can be unset

iLLiCiTiT:
## Changelog Description
Alllow `handleStart` and `handleEnd` to be `None` on version.

Attributes `handleStart` and `handleEnd` are always filled but can be filled with `None`. The handling method `get` would be working only if the attribute would not be available on version. The change works for both `None` and `0`.

## Testing notes:
1. Plates without filled `handleStart` or `handleEnd` filled on version can be loaded.


BigRoy (approved):
Admittedly - this may fix the error that happened. However, I still think it points to an API bug or server bug because a number attribute should not be allowed to be `None` explicitly as a value I'd say.

PR #133, opened 2025-07-22
Publish: Use new workfiles API to increment current workfile.

BigRoy:
## Changelog Description

When incrementing the workfile during publish, use the new context-based workfile API so that the current author information is stored with the saved file.

Requires latest develop `ayon-core` with this PR included: https://github.com/ynput/ayon-core/pull/1275

## Testing notes:

1. Publishing should work
2. Workfile should be incremented
3. Workfile should show the author of the publish 
4. Workfile should have 'artist note' set: "Incremented by publishing."
5. Publishing should also still work and increment (the old way) with older ayon-core.

### DIFF
--- a/client/ayon_nuke/plugins/create/create_backdrop.py
+++ b/client/ayon_nuke/plugins/create/create_backdrop.py
@@ -16,7 +16,6 @@ class CreateBackdrop(NukeCreator):
     label = "Nukenodes (backdrop)"
     product_type = "nukenodes"
     icon = "file-archive-o"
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/api/lib.py
```diff
@@ -928,8 +933,11 @@ def get_version_from_path(file):
 def version_up_script():
```
iLLiCiTiT, line 933: It looks like this function is not used at all...

PR #169, opened 2025-12-11
LoadClip: Add auto_duration checkbox

vincentullmann:
## Changelog Description
Adds an `auto_duration` checkbox to the `LoadClip` loader which, if enabled will read the source medias clip duration

- does not yet work for filesequences

## Testing notes:
1. publish a video file
2. edit the product metadata to generate a mismatch between the stored duration and the actual video file duration
3. import without the checkbox: the clip should use the metadatas duration
4. import with the checkbox: the clip should use the files duration

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -16,6 +16,7 @@ from ayon_nuke.api.lib import (
     get_imageio_input_colorspace,
     maintained_selection
 )
+from ayon_core.plugins.load.export_otio import get_image_info_metadata
 from ayon_nuke.api import (
     containerise,
     update_container,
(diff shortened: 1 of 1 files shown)


BigRoy:
Is this just trying to fix the fact that your published product _actually_ has the wrong metadata on publish @vincentullmann ?
It seems a bit weird to start deviating on purpose from our stored metadata. Is the issue not just that we're storing it incorrectly to begin with?

BigRoy:
@vincentullmann could you see if https://github.com/ynput/ayon-nuke/pull/239 would solve your needs - there you could just disable AYON taking control of the frame range and instead leave it up to Nuke itself to figure it out at that point?

vincentullmann:
@BigRoy did a quick test and yes it does seem to work. thank you!

Issue #155, opened 2025-10-31
YN-0377 - Collect Writes fails if there are more leading zeroes in file sequence


jrsndl:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Currently, when publishing with slate for a first time, and templates frame padding is set higher then number of digits of the start frame, Collect Write fails.

### Steps To Reproduce:

1. Frame padding set to 5
<img width="619" height="167" alt="Image" src="https://github.com/user-attachments/assets/5e116255-b5a4-4ebd-a791-e6ca33daa479" />
2. Publish from Nuke, shot range 1001-1010, slate 1000
3. Collect Writes validator fails, 'cause frame_start_index is None

### Version
server 1.10.1, ayon-core 1.6.1, nuke 0.3.12

### What platform were you running when you found the bug?
Windows 10 

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell
Traceback (most recent call last):
File "C:\Users\john.doe\AppData\Local\Ynput\AYON\dependency_packages\ayon_2502101448_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_processrunner(*args)
File "C:\Users\john.doe\AppData\Local\Ynput\AYON\addons\nuke_0.3.12\ayon_nuke\plugins\publish\collect_writes.py", line 53, in process self._set_existing_files_data(instance, colorspace)
File "C:\Users\john.doe\AppData\Local\Ynput\AYON\addons\nuke_0.3.12\ayon_nuke\plugins\publish\collect_writes.py", line 81, in _set_existing_files_data representation = self._get_existing_frames_representation(
File "C:\Users\john.doe\AppData\Local\Ynput\AYON\addons\nuke_0.3.12\ayon_nuke\plugins\publish\collect_writes.py", line 288, in _get_existing_frames_representation collected_frames = self._add_slate_frame_to_collected_frames(
File "C:\Users\john.doe\AppData\Local\Ynput\AYON\addons\nuke_0.3.12\ayon_nuke\plugins\publish\collect_writes.py", line 383, in _add_slate_frame_to_collected_frames
slate_frame = collected_frames[frame_start_index].replace(
TypeError: list indices must be integers or slices, not NoneType
```

[YN-0377](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=28d890d0f06511f0a4e3c7a960803aed)

PR #173, opened 2026-01-13
Collect Writes should allow more values in frame padding 

moonyuet:
## Changelog Description
This PR is to fix the issue where Collect Writes fails if there are more leading zeroes in file sequence.
Fixes: https://github.com/ynput/ayon-nuke/issues/155

n/a

## Testing notes:
1. Create pre-render or any render related instance with the slate frame
https://github.com/ynput/ayon-documentation/blob/d3eb215955d8309ccc788ee381271d1ef804a450/website/docs/addon_nuke_admin.md#slate-workflow2. Publish


jakubjezek001 (approved):
looking good. Gosh I dont get the complexity of the old code (me years ago, right :D )

PR #175, opened 2026-01-15
fix: remove ' from product_base_type

vincentullmann:
## Changelog Description
Removed an additional `'` on the `product_base_type` value

Error Message:
```
        },
        "folderId": "80d77490b98811f09b870a1e7b16a38b",
        "productBaseType": "model'"
    }
}
Detail: Invalid data provided: 1 validation error for ProductPostModel
productBaseType
  string does not match regex "^[a-zA-Z0-9_]([a-zA-Z0-9_\.\-]*[a-zA-Z0-9_])?$" (type=value_error.str.regex; pattern=^[a-zA-Z0-9_]([a-zA-Z0-9_\.\-]*[a-zA-Z0-9_])?$).
 ```

### DIFF
--- a/client/ayon_nuke/plugins/create/create_model.py
+++ b/client/ayon_nuke/plugins/create/create_model.py
@@ -14,7 +14,7 @@ class CreateModel(NukeCreator):
     identifier = "create_model"
     label = "Model (3d)"
     product_type = "model"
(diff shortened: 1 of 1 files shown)


BigRoy:
Thanks @vincentullmann - hotfix release will be out in a bit.

Issue #162, opened 2025-11-19
Load Clip Plugin should handle missing frameStart/frameEnd version attribute


TobiasPharos:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
https://github.com/ynput/ayon-nuke/blob/c148e7ec47b914fbcecfd4f646ec8752537d26bc/client/ayon_nuke/plugins/load/load_clip.py#L135C1-L138C32

currently sets `first` and/or `last` to None if the version attributes `frameStart` or `frameEnd` aren't set, which then leads to an error in the following calculation, which tries to add/subtract a number to/from None.

## Expected Behavior
Should fail gracefully or have a fallback

### Steps To Reproduce:

Load a version without frameStart set using Load Clip.
Loading via the Ayon Loader leads to a cryptic error, loading via a workfile template will fail to resolve the corresponding placeholder.

### Relevant log output:

```shell
TypeError: unsupported operand type(s) for -=: 'NoneType' and 'int'
```

PR #176, opened 2026-01-16
Load Clip: handling missing frameStart/frameEnd version attribute 

moonyuet:
## Changelog Description
This PR is to ensure the frameStart/frameEnd version attributes should be handle in Clip loader when they are missing. With this PR, the TypeError contributes to these missing attributes would be resolved.

Resolve https://github.com/ynput/ayon-nuke/issues/162

n/a

## Testing notes:
1. Launch Nuke
2. Load Clip and Setting Version of it.

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -152,10 +152,6 @@ class LoadClip(plugin.NukeLoader):
         # but file sequences its the first frame that is 1 frame lower.
         slate_frames = repre_entity["data"].get("slateFrames", 0)
         extension = "." + repre_entity["context"]["ext"]
-
-        if extension in VIDEO_EXTENSIONS:
-        slate_frames = repre_entity["data"].get("slateFrames", 0)
-        extension = "." + repre_entity["context"]["ext"]
         files_count = len(repre_entity["files"])
 
         if first is not None and last is not None:


client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -137,8 +137,16 @@ def load(self, context, name, namespace, options):
         self.handle_start = version_attributes.get("handleStart") or 0
         self.handle_end = version_attributes.get("handleEnd") or 0
```
BigRoy: Can you remove the `self.` here please. No idea why that's there.

client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -309,8 +317,16 @@ def update(self, container, context):
...
```
BigRoy: Can we remove the `self.` here?

client/ayon_nuke/plugins/load/load_clip.py
rdelillo: If `first` or `last` is None, some of the following code fails with `TypeError`.

```suggestion
        slate_frames = repre_entity["data"].get("slateFrames", 0)
        extension = "." + repre_entity["context"]["ext"]
        files_count = len(repre_entity["files"])

        if first is not None and last is not None:
            if not is_sequence:
                duration = last - first
                first = 1
                last = first + duration

            # If a slate is present, the frame range is 1 frame longer for movies,
            # but file sequences its the first frame that is 1 frame lower.
            if extension in VIDEO_EXTENSIONS:
                last += slate_frames
            elif extension in IMAGE_EXTENSIONS and files_count != 1:
                first -= slate_frames
```

Issue #105, opened 2025-04-14
AY-7696_Loaded "Import Nuke Nodes" backdrops can't be updated


ynbot:
### Please describe the issue you have encountered?

When doing "Import Nuke Nodes" on e.g. silhouette roto shapes it will create the nodes on a backdrop, which however fail to update. 

On update it seemed to delete the backdrop and reimport new node - so now you have 'duplicates' which seem to be the old AND the new nodes?

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1326234407908020274/1359127236262756436)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7696](https://app.clickup.com/t/86c2zypyx)_

PR #177, opened 2026-01-20
Import Backdrop Node should remove previous node during update

moonyuet:
## Changelog Description
This PR is to make sure importing backdrop node should remove previous nodes during up-versioning or setting version, and it also add options to allow users to choose whether they would like to remove all backdrop nodes during removing instances by enabling `ayon+settings://nuke/load/LoadBackdropNodes/remove_all_backdrops_nodes`

Resolve https://github.com/ynput/ayon-nuke/issues/105

Should we use the latest version of `getting_ayon_data` alike instead of `get_avalon_data`?

## Testing notes:
1. Load backdrop node
2. Update/Set Version

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -673,6 +673,48 @@ def get_nuke_imageio_settings():
     return get_project_settings(Context.project_name)["nuke"]["imageio"]
 
 
+def get_matching_override_node(node_class, plugin_name, product_name):
+    """Find matching override node for the given configuration.
+
+    Args:
+        node_class (str): Nuke node class name
+        plugin_name (str): Plugin name
+        product_name (str): Product name
+
+    Returns:
+        dict or None: Matching override node or None if not found
+    """
+    imageio_nodes = get_nuke_imageio_settings()["nodes"]
+    override_nodes = imageio_nodes["override_nodes"]
+
+    for override_node in override_nodes:
+        node_class_preset = override_node["nuke_node_class"]
+
+        if override_node.get("custom_class"):
+            node_class_preset = override_node["custom_class"]
+
+        if node_class not in node_class_preset:
+            continue
+
+        if plugin_name not in override_node["plugins"]:
+            continue
+
+        product_names = override_node["product_names"]
+
+        if (
+            product_names
+            and not any(
+                re.search(s.lower(), product_name.lower())
+                for s in product_names
+            )
+        ):
+            continue
+
+        return override_node
+
+    return None
+
+
 def get_imageio_node_setting(node_class, plugin_name, product_name):
     ''' Get preset data for dataflow (fileType, compression, bitDepth)
     '''
@@ -694,13 +736,24 @@ def get_imageio_node_setting(node_class, plugin_name, product_name):
     if not imageio_node:
         return
 
-    # find overrides and update knobs with them
-    get_imageio_node_override_setting(
-        node_class,
-        plugin_name,
-        product_name,
-        imageio_node["knobs"]
+    # Check if a node override exists for this configuration
+    override_imageio_node = get_matching_override_node(
+        node_class, plugin_name, product_name
     )
+
+    # If node override exists,
+    # use only override knobs
+    # (don't merge with original master knobs)
+    if override_imageio_node:
+        imageio_node["knobs"] = override_imageio_node.get("knobs", [])
+    else:
+        # Otherwise apply partial overrides to original master knobs
+        imageio_node["knobs"] = get_imageio_node_override_setting(
+            node_class,
+            plugin_name,
+            product_name,
+            imageio_node["knobs"]
+        )
     return imageio_node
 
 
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/plugins/load/load_backdrop.py
```diff
@@ -145,7 +146,7 @@ def load(self, context, name, namespace, data):
                     continue
                 else:
                     new_nodes.append(n)
-
+            data_imprint["nodes"] = json.dumps([node.name() for node in new_nodes])
```
BigRoy: I don't think this safe - if I copy paste a loaded backdrop with all the nodes then the duplicate will share the same data - which _should_ show in the Scene Inventory too. Then removing or updating from scene inventory will now delete the original nodes by name, not the duplicates.

Can we rely on something here that is not node names that may be 'changed' over time? Can we rely somehow on what actually "belongs" to the backdrop? Since we're managing a backdrop, can't we remove whatever is in/on it?

BigRoy: I think you misunderstood @moonyuet - you're still imprinting some data into nodes - that are nodes or ids that are actually incorrect.

Say I store that my node id is X and I have this backdrop node with all its nodes inside, I duplicate them manually inside the graph - both have now child nodes stored with that id X... hence on removal, which should I remove.

What I said is, what if when we remove a backdrop, we just also delete all its child nodes?
See [`nuke.BackdropNode.getNodes()`](https://learn.foundry.com/nuke/developers/14.0/pythonreference/_autosummary/nuke.BackdropNode.html#nuke.BackdropNode.getNodes)?

Why can't we just delete all the nodes belonging with the backdrop?

server/settings/loader_plugins.py
```diff
@@ -48,6 +49,13 @@ class LoadClipOptionsModel(BaseSettingsModel):
         default="auto",
     )
 
+
+class LoadBackdropNodesModel(BaseSettingsModel):
+    remove_nodes_from_backdrop: bool = SettingsField(
+        title="Remove existing Ayon backdrops when removing container"
```
BigRoy:
```suggestion
        title="Remove existing AYON backdrops when removing container"
```

client/ayon_nuke/plugins/load/load_backdrop.py
```diff
@@ -219,15 +201,14 @@ def update(self, container, context):
             xpos = GN.xpos()
             ypos = GN.ypos()
             avalon_data = get_avalon_knob_data(GN)
+            for node in GN.getNodes():
+                nuke.delete(node)
             nuke.delete(GN)
             # add group from nk
             nuke.nodePaste(file)
```
BigRoy: Doesn't this replacement somehow need logic also to preserve incoming and outgoing connections? It feels like now we're just deleting the old and generating the new but that would essentially make it lose any connections I suppose?

moonyuet: Actually, using the old one would lose the ability to remove the nodes outside the backdrop but generating the new one won't.

client/ayon_nuke/plugins/load/load_backdrop.py
```diff
@@ -219,15 +201,60 @@ def update(self, container, context):
...
+
+                # Outgoing connections of the node
+                for dependent in node.dependent():
+                    for input_index, depcy in enumerate(dependent.dependencies()):
+                        if node is depcy:
+                            node_connections.append({
+                                "node_name": node.name(),
+                                "dependent_name": dependent.name(),
+                                "dependent_input_index": input_index
+                            })
```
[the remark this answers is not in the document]
moonyuet: I tried to store it as object but for some reasons there would be a PythonObject Error raised due to nuke would convert the node object into str. I have mentioned in the docstring but I guess we can discuss the best way to resolve PythonObject.

rdelillo:
Not sure this is ready just yet, I gave it a try tonight, it ended up very similarily with what @jakubjezek001 posted.

IMO the connection restore thing should only register/restore the connection from the outside to the inside and the other way around but not the connections within the backdrop itself. 
The DOT nodes connections don't seem to be restored properly but might not be related to this PR, and something on how the backdrop are published.

I've also experienced the Python knob error messag, but not sure where this comes from.

BigRoy:
> IMO the connection restore thing should only register/restore the connection from the outside to the inside and the other way around but not the connections within the backdrop itself.
> The DOT nodes connections don't seem to be restored properly but might not be related to this PR, and something on how the backdrop are published.

This ☝️ - it should not affect anything within the backdrop, especially because on update new nodes may exist entirely - even with intentionally new connections, etc.

moonyuet:
> WARNING:SceneInventory:Update failed
>   File "C:\CODE\__YNPUT\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 1120, in _update_containers
>     update_container(container, item_version)
>   File "C:\CODE\__YNPUT\ayon-core\client\ayon_core\pipeline\load\utils.py", line 571, in update_container
>     return Loader().update(container, context)
>   File "C:\CODE\__YNPUT\ayon-core\client\ayon_core\pipeline\load\plugins.py", line 426, in wrapped_method
>     result = original_method(self, *args, **kwargs)
>   File "C:\CODE\__YNPUT\ayon-nuke\client\ayon_nuke\plugins\load\load_backdrop.py", line 201, in update
>     with maintained_selection():
>   File "C:\Program Files\Nuke15.0v4\python310.zip\contextlib.py", line 142, in __exit__
>     next(self.gen)
>   File "C:\CODE\__YNPUT\ayon-nuke\client\ayon_nuke\api\lib.py", line 2356, in maintained_selection
>     select_nodes(previous_selection)
>   File "C:\CODE\__YNPUT\ayon-nuke\client\ayon_nuke\api\lib.py", line 2420, in select_nodes
>     node["selected"].setValue(True)
> ValueError: A PythonObject is not attached to a node
> ```

Fixed in [8df4d3b](https://github.com/ynput/ayon-nuke/pull/177/commits/8df4d3b8e97e7b23d8c71c5fc5c69aa719e7d979)

PR #180, opened 2026-01-26
🏛️Product base types: Make use of product base

### LINKED ISSUE #179: 🏛️Product base types: Make use of product base types
## Feature
Once all support is merged in `ayon-core`, change how Nuke addon is working with product base types:
- all creators should have `product_base_type` defined
- settings for creators exist so you can define product types per creator
- all setting profiles/filters are changed so current product type changes into product base type along with the conversion of existing settings
- pyblish plugins that are using either profiles or `family` == `product_type` are changed to make use of `product_base_type`
- depencency on **ayon-core** implementing all necessary functionality is defined in `package.py`

antirotor:
## Changelog Description
Add support for product base types. Change the current usage of product type, add settings and conversion methods.

This PR requires future version of ynput/ayon-core with proper support (most likely 1.9.0)

## Testing notes:
Without compatible version of ayon-core, this won't work. With it, everything should work as before plus it should set properly `product_type` if set using creator settings.

Closes #179

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -673,6 +673,48 @@ def get_nuke_imageio_settings():
     return get_project_settings(Context.project_name)["nuke"]["imageio"]
 
 
+def get_matching_override_node(node_class, plugin_name, product_name):
+    """Find matching override node for the given configuration.
+
+    Args:
+        node_class (str): Nuke node class name
+        plugin_name (str): Plugin name
+        product_name (str): Product name
+
+    Returns:
+        dict or None: Matching override node or None if not found
+    """
+    imageio_nodes = get_nuke_imageio_settings()["nodes"]
+    override_nodes = imageio_nodes["override_nodes"]
+
+    for override_node in override_nodes:
+        node_class_preset = override_node["nuke_node_class"]
+
+        if override_node.get("custom_class"):
+            node_class_preset = override_node["custom_class"]
+
+        if node_class not in node_class_preset:
+            continue
+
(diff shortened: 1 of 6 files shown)


server/settings/create_plugins.py
```diff
@@ -92,6 +92,16 @@ def ensure_unique_names(cls, value):
         ensure_unique_names(value)
         return value
 
+class ProductTypeItemModel(BaseSettingsModel):
+    product_type: str = SettingsField(
```
BigRoy: This looks much better when using compact layout.
```suggestion

class ProductTypeItemModel(BaseSettingsModel):
    _layout = "compact"
    product_type: str = SettingsField(
```

iLLiCiTiT:
> Currently with the latest core -> when plates are published with CSV_ingest, AYONLoader isn't showing Loadclip option in nuke

Hi, is that related to this PR?

server/settings/conversion.py
```diff
@@ -232,6 +232,25 @@ def _convert_baking_stream_filter_product_base_type_0_4_9(
...
+    """Convert collect instance data model to include product_base_type."""
+    if "publish" not in overrides:
+        return
+
+    if not overrides["publish"].get("CollectInstanceData"):
+        return
+
+    collect_instance_data = overrides["publish"]["CollectInstanceData"]
+    if not collect_instance_data:
+        return
```
BigRoy: This will never happen because above you're already checking for a value.

I'd just do:
```python
    collect_instance_data = overrides["publish"].get(
        "CollectInstanceData"
    )
    if not collect_instance_data:
        return
```
Right?

Issue #68, opened 2025-02-07
AY-7439_Node override validation mismatch


ynbot:
### Please describe the issue you have encountered?

If a node override triggered by product variant presets is used, the render write node validation should not validate knob values based on the overridden values. Currently, the validator merges the original knob presets with the overridden knob presets. In some cases, this causes the validator to crash because some knob values are not available under the overridden variants.

Best approach here would be to avoid original master knob variants if node override preset is used.

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7439](https://app.clickup.com/t/86c1yw59k)_

PR #189, opened 2026-02-04
Uses override knobs only when override is defined in settings

moonyuet:
## Changelog Description
This PR is to make sure override knobs is used only if override knobs setting has been defined. 
Resolve https://github.com/ynput/ayon-nuke/issues/68

Need to also test without any override knobs

## Testing notes:
1. Define some override knobs in `ayon+settings://nuke/imageio/nodes/override_nodes`
2. Create Render
3. Publish
4. Should be published successfully 
5. Clean up some nodes
6. Publish
7. Validation blocks but once hit the repair action and restart publishing, it shouldn't be errored out anymore again.

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -143,11 +143,6 @@ class LoadClip(plugin.NukeLoader):
             version_attributes, handle_start, handle_end
         )
 
-        if first is not None and last is not None and not is_sequence:
-            duration = last - first
-            first = 1
-            last = first + duration
-
         # If a slate is present, the frame range is 1 frame longer for movies,
         # but file sequences its the first frame that is 1 frame lower.
         slate_frames = repre_entity["data"].get("slateFrames", 0)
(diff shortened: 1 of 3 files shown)


client/ayon_nuke/api/lib.py
```diff
@@ -793,7 +819,7 @@ def get_view_process_node():
     ipn_node = None
     for v_ in nuke.allNodes(filter="Viewer"):
         ipn = v_['input_process_node'].getValue()
-        ipn_node = nuke.toNode(ipn)
+        ipn_node = nuke.toverride_node(ipn)
```
BigRoy: Pretty sure these changes are wrong @moonyuet ? :) Search-replaced a bit too much perhaps?


```suggestion
        ipn_node = nuke.toNode(ipn)
```

client/ayon_nuke/api/lib.py
```diff
@@ -1957,7 +1983,7 @@ def set_reads_colorspace(self, read_clrs_inputs):
 
             if nuke.ask(msg):
                 for nname, knobs in changes.items():
-                    n = nuke.toNode(nname)
+                    n = nuke.toverride_node(nname)
```
BigRoy:
```suggestion
                    n = nuke.toNode(nname)
```

PR #193, opened 2026-02-11
Added setting for disabling forcing context settings

splidje:
## Changelog Description
Added setting for disabling forcing context settings (e.g. frame range) upon opening an existing Script.

Added a new setting under general described as "Set Context Settings on Script Open".
Defaults to True, but if False opening a Nuke Script by any means - e.g. via Ayon -> Workfiles, selecting from the Launcher, using vanilla Nuke Open - skips calling `set_context_settings`, so e.g. keeps Root node first/last frame as it was on save.

## Testing notes:
1. Create a new blank Nuke Script
2. Set frame range to something different from the context's frame range (except not 1-100, which is Nuke's default and it won't save into the .nk)
3. Save
4. Close Nuke
5. Open the script again
6. Frame range should be the same as it was when it was saved


client/ayon_nuke/api/lib.py
```diff
@@ -819,6 +819,25 @@ def get_view_process_node():
         return duplicate_node(ipn_node)
 
 
+def on_create_root():
+    if script_name():
+        log.info("Not a new Script. Skipping setup.")
+        return
+
+    log.info("New Script. Setting up ...")
```
BigRoy: This seems like new logic - and seems like logic to differentiate between "scene open" and "new scene" if I'm not mistaken? However, if that's the case shouldn't the _settings based toggle_ on whether it should affect upon opening an existing script happen here.

Because the way it's done now makes it feel a bit like it's **also** skipping it now for New Scripts if the setting is disabled because this whole callback does not get registered.

BigRoy (commented):
Very nice - disabling the integration to mess with your scenes on open I think is a good thing. My personal preference, I don't want to open your scene and have it altered without your knowledge. So this is a step the good way.

I'd love to still be notified and asked if I want to fix certain things though, but not super important.

@jakubjezek001 @LiborBatek thoughts?

client/ayon_nuke/api/lib.py
BigRoy: Actually I think I've misread and this is actually still always registered 👍 if so, then this comment still applies: https://github.com/ynput/ayon-nuke/pull/193#discussion_r2795875068

nebukadhezer:
But isnt that exactly what a pipeline should do set the correct frame range ? what is the use case ? under a task with a defined frame range you dont want to have that frame range ? Or maybe I am not getting what this is ment for...
But if egg editorial changes frame ranges on a running production I want those ranges to propagate, does it need to be on startup ? No, easy on publish with validation plus fix method...

splidje:
My particular use case is an artist working under a test shot doing various experiments.

In my previous job we did consider it very important to ensure people are working with the correct context settings, however we also didn't like to impose too much in the way of changing someone's work file automatically without flagging the changes.

We had a "set the correct context settings" menu item so an artist could ensure it's all correct, and also various moments where they'd be checked and if wrong a dialog would pop up telling them which were wrong and asking whether to set them correctly or not.

Then obviously on creating a new work file, those settings would be set correctly.


I guess another example might be the frame changing after a work file has been created. The artist does need to have their frame changed, but they need to be notified about it before it happens as they might want to sort a few things out before setting it.

nebukadhezer:
I agree that user notification is a good thing, a validator on publish does that... I still think that I lean more towards maybe an "opt out" button per script, rather than a studio wide setting ? Would be nice to have some metadata stored on root and if that says "dont update me" the callback just skips the update ? Just thinking out loud ... but I do get the usecase ... that if you are testing stuff to strict pipelines are not helpful...

splidje:
Random question/suggestion:

I've got into the habit over the last few years of running Black on Python code to 1) have a consistent formatting and 2) not have to really think about formatting.

Reckon something like that might be a nice idea for these repos?

client/ayon_nuke/api/pipeline.py
```diff
@@ -181,8 +179,9 @@ def add_nuke_callbacks(project_settings: dict = None):
     nuke.addOnScriptLoad(check_inventory_versions)
     nuke.addOnScriptSave(check_inventory_versions)
 
-    # set apply all workfile settings on script load and save
-    nuke.addOnScriptLoad(WorkfileSettings().set_context_settings)
+    if nuke_settings["general"]["set_context_settings_on_script_open"]:
+        # set apply all workfile settings on script load and save
```
BigRoy: Pretty sure this comment above it was just wrong?
```suggestion
        # set apply all workfile settings on script load
```

splidje:
I've made those tweaks you pointed out, and also put in asking for confirmation, changing the general setting to "check" rather than "set" and added support for a specific script to have checking turned off.

client/ayon_nuke/api/lib.py
```diff
@@ -2062,22 +2127,64 @@ def reset_frame_range_handles(self):
...
+                        # update node graph so knobs are updated
+                        update_node_graph()
+                        was_node_graph_updated = True
+                    for knob_name, _, new_value in knob_name_current_new_value_triplets:
+                        node[knob_name].setValue(new_value)
+
+                # in case no Viewers needed knob changes
+                if not was_node_graph_updated:
+                    # update node graph so knobs are updated
+                    update_node_graph()
```
BigRoy: Why does this suddenly need all this complexity? This is very hard to read and grasp.

Admittedly - all we're really changing is:
- Before `reset_frame_range_handles` gets called, ask for confirmation of changes.
- Or better, before any of the `reset_..._` gets called handle confirmation.

To know whether we need to ask the user we need two things:
- 1. Are we checking?
- 2. Is there anything to change?

The way I see it is what we should do is separate the logic of getting the current and expected state from the setting the state.

```python
# overly simplified psuedocode
should_check = ...[settings][check][blabla]
if not should_check:
    return

current, expected = get_current_state()
if current == expected:
    return

if headless:
    # log difference
    log.warning("Values mismatch: {current} != {headless}")
    return

accepted = prompt_user(current, expected):
if accepted:
    set_state(expected)
```

But in short, the setting of the values should not be changing or get more complicated at all?
We just end up having a library `set_frame_range`, `set_colorspace` and `set_frame_range` functions that will end up being called, right?

BigRoy (changes requested):
Thanks - I think code got worse now with latest changes - so more tweaks will be needed. I'll comment my first pass - @jakubjezek001 who's off this week may be better suited to comment from Nuke's perspective so my comments are mainly on the coded approach, not anything nuke-specific.

jakubjezek001:
> Back to this. How do I use ruff? I can see a `ruff.toml` file there. Is it similar to `black` in that I can just run `ruff .` when in the repo folder?

try `ruff check .` inside the addon root dir. But you need to be in venv with installed ruff.

moonyuet:
I tested the PR by enabling `ayon+settings://nuke/general/check_context_settings_on_script_open`. When switching the task, it would pop up dialog to ask you whether you would like to change the colorspace. When it turns up it doesn't show the dialog.
image

We can introduce confirmation dialog but make sure it is not executed when it is at headless mode.

PR #195, opened 2026-02-19
Added creator for plates

splidje:
## Changelog Description
For my purposes, I can happily just use Nuke to publish plates by converting sections of MXFs to EXR.
So it made sense to add a creator for the plate product type.


splidje:
That's accurate, yeah.

One main requirement is just "aesthetic" (probably not the right word), i.e. just eye-balling a product and seeing that it's a plate.

But I'll likely have some logic at some point which benefits from knowing what is a plate and what isn't.

Other than that, what I have set up atm is that writing out a plate is very similar to a render. Just with a few configuration differences.

PR #201, opened 2026-02-25
YN-0526 - Feat/allow different write node types

### LINKED ISSUE #202: YN-0526: Support for custom Write Nodes

we'd like to add a few custom features to the write nodes in nuke (some additional buttons and knobs) One way to do this, would be to setup a custom gizmo, eg.: stray_write which we could then use as our default inhouse write node. many of the key knobs would have matching names to the stock write node (for artists sake) but also to allow code written for the stock Write node to support it. Here is a [draft PR](https://github.com/ynput/ayon-nuke/pull/201) where I started making some changes, how I'd approach supporting this. Its still work in progress but I'm sharing it now so we can discuss the overall idea/direction before going into implementation details Is this supporting custom write nodes something we could support in ayon or does it conflict with any other concepts? 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1476188109140922551/1476188115088572477)

_This issue was automatically created from AYON ticket [YN-0526]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=e5680840125011f186e58b4ed7c4918b)

vincentullmann:
# WIP: Not ready to be merged yet

## Changelog Description
Refactored Collection and Extraction of Write Nodes to allow for easier replacement of individual steps.

- **Collect Writes**: Refactored to only collect as little data as possible -> making it easier to implement other `Collect<NodeType>` Steps
   Currently collects:
     - `node`: the main / parent node
     - `write_node`: the node that will be generating the output (can be the same as "node")
     - `frame_range`: tbd: might be a candidate to be moved out too
     - `expected files`
- **Collect Color Space**: New plugin that collects colorspace (refactored out of "Collect Writes")
- **Validate Write Nodes**: updated to use the new structure
- **Validate Output Resolution**: updated to use the new structure
- **NukeRenderLocal**: simplified to basically just run `nuke.execute(node, frame_ranges)`.
  - added a bit of logic to check for optional `instance.data["frames_to_render"]`
  - :wastebasket: removed "frames to fix" / last-published copy logic and representation building (to be handled in other steps).
- **Collect Existing Frames** new place for the "copy last-published logic"
  [ :warning: not implemented yet.]
  Planned features:
  - check for prev. publish
  - copy files
  - update `instance.data["frames_to_render"]`

## Further Steps / Ideas:
- implement "Collect Existing Frames"
- introduce node_type based families
  - similar to how `ayon-houdini` sets families both: based on productType (eg.: "render") as well as node_type (eg.: "arnold_rop")
   - this would allow us to more specifically target node_types in certain steps, allowing developers to implement specific plugins for individual node_types
   

## Testing notes:
1. start with this step
2. follow this step

---

[YN-0526](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=e5680840125011f186e58b4ed7c4918b)
Fix https://github.com/ynput/ayon-nuke/issues/202

### DIFF
--- a/client/ayon_nuke/api/__init__.py
+++ b/client/ayon_nuke/api/__init__.py
@@ -13,9 +13,10 @@ from .plugin import (
     NukeCreator,
     NukeWriteCreator,
     NukeCreatorError,
+    find_knob_on_node,
+    find_node_with_knob,
     get_instance_group_node_childs,
     get_instance_group_node_children,
-    get_colorspace_from_node,
     get_publish_config,
 )
 from .pipeline import (
@@ -71,9 +72,11 @@ __all__ = (
     "NukeWriteCreator",
     "NukeCreatorError",
     "NukeHost",
+    "find_knob_on_node",
+    "find_node_with_knob",
     "get_instance_group_node_childs",
     "get_instance_group_node_children",
-    "get_colorspace_from_node",
+
     "get_publish_config",
 
     "ls",
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -589,15 +589,60 @@ def get_instance_group_node_childs(instance):
     return get_instance_group_node_children(instance)
 
 
-def get_colorspace_from_node(node):
-    # Add version data to instance
-    colorspace = node["colorspace"].value()
+# should this be in lib.py?
+def find_knob_on_node(
+    node: nuke.Node,
+    knob_names: list[str],
+    *,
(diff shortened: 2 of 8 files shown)


vincentullmann:
closing as its no longer required.

I've replaced `NukeRenderLocal` with a few smaller plugins in our inhouse plugin, that work for both default and custom write nodes (assuming that key knobs such as `file`, `colorspace`, etc match)

PR #204, opened 2026-02-26
Plate reads timewarp

splidje:
## Changelog Description
Pulls timewarp data from a TimeWarp node downstream of the create node when the product type is plate (for use with turning over plates using Nuke)

Carrying on from here: https://github.com/ynput/ayon-nuke/pull/203

### DIFF
--- a/client/ayon_nuke/api/colorspace.py
+++ b/client/ayon_nuke/api/colorspace.py
@@ -201,6 +201,7 @@ def get_formatted_display_and_view(
         # making sure formatted colorspace exists in running session
         if dv_item_resolved in get_display_and_view_colorspaces(root_node):
             return dv_item_resolved
+    return ""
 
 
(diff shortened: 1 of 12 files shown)


jakubjezek001:
I was investigating even more and it seems the plotting is not that precise.. 
image

It is not a big issue in this particular examle but somewhere the approximation might be the problem.

iLLiCiTiT (changes requested):
This is based on custom product types, they ~~were~~ are not meant for core pipeline functionality.

splidje:
> This is based on custom product types, they ~were~ are not meant for core pipeline functionality.

Does that mean for this to be viable I'd need to add a creator for product base type "plate" similar to this:
https://github.com/ynput/ayon-nuke/pull/203

And therefore have the `if` statement test `product_base_type` rather than `product_type`?

PR #209, opened 2026-03-06
Fix for being able to load in effects published from hiero in nuke again

robertokker:
## Changelog Description
added plate in the product_base_type again otherwise you can't load in the effects

image


robertokker:
> I'm a bit confused here - where do these "effects" get published from? There is no `effect` product type whatsoever that can be exported from Nuke integration. So I guess it's from Hiero integration?
> 
> I feel the issue is that Hiero is publishing these as `plate` base product type to begin with. The issue is not on the loader, but on the published product.
> 
> @jakubjezek001 can you confirm how this _should_ work? @iLLiCiTiT do we have a backwards compatible bug for Hiero?

the effects are getting published from hiero and are always connected to the plate.
You can't publish only effects.

https://github.com/ynput/ayon-hiero/pull/110

PR #221, opened 2026-03-19
Added creator for product base type "plate"

splidje:
## Changelog Description
I've decided to use just plain Nuke to turnover plates from Editorial using this studio specific addon:
https://github.com/splidje/ayon-ubp-nuke-turnover

For it I need a create instance for the product type "plate".

I've also reduced some duplicated code by combining: `CreateWriteRenderModel` `CreateWritePrerenderModel` and `CreateWriteImageModel` into one general base class `CreateWriteModel`


jakubjezek001 (approved):
Works well

BigRoy (changes requested):
> For it I need a create instance for the product type "plate".

Why did you need this product base type? Please elaborate? I really don't see why we need yet another Creator that just happens to define a `product type` which should really just be an alias to the product base type? Why does it absolutely require to be its own product base type? Please elaborate.

Code looks absolutely fine. I'm just confused WHY we'd want to duplicate it like this.

This looks very much like duplicates to:
- https://github.com/ynput/ayon-nuke/pull/203
- https://github.com/ynput/ayon-nuke/pull/195

Which both were not merged for that same reason.

PR #234, opened 2026-04-08
Fix/render on farm button

### LINKED ISSUE #167: Render On Farm button should submit only itself to farm

## Current Behavior
Render On Farm button on write node should send only the selected write node, ignoring the other write nodes which are turned on in the publish window

## Expected Behavior
other write nodes shouldnot get submit to deadline(without going to publish window again and manually turning off)

### Steps To Reproduce:

create multiple write nodes in the scene
Hit renderonfarm for one write node
should submit only the selected write node

### Version
Version of the server, ayon-core and the addon that's affected

Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

semx2a:
Changelog Description

Fix "Render On Farm" button submits all active write nodes instead of only the clicked one.

When clicking "Render On Farm" on a write node, the old code only sets `active=True` for the matching instance without deactivating the others. `CollectFromCreateContext` (order `CollectorOrder - 0.5`) collects all `active=True` instances before `CollectRenderOnFarm` (order `CollectorOrder - 0.49`) can deactivate them, so any write node already enabled in the publish window would also get submitted to the farm.

The fix explicitly sets `active=False` for all non-matching instances in `_submit_render_on_farm` before pyblish runs, so they are never collected into the context.

Closes #167

## Testing notes:

1. Create two or more write nodes with `render_on_farm` enabled in instance attributes.
2. Ensure all write nodes are active in the publish window, then close it.
3. Click **Render On Farm** on one write node — verify only that node appears as a job in Deadline.
4. Repeat for a different write node and confirm the same isolation behaviour.
5. Verify a normal publish (via the publish window with multiple nodes active) is unaffected — all active nodes still publish as expected.

### DIFF
--- a/client/ayon_nuke/api/utils.py
+++ b/client/ayon_nuke/api/utils.py
@@ -125,12 +125,11 @@ def _submit_render_on_farm(node) -> bool:
     host = registered_host()
     create_context = CreateContext(host)
 
-    # Ensure CreateInstance is enabled.
+    # Ensure only the target CreateInstance is enabled, deactivate all others.
     for instance in create_context.instances:
-        if node.name() != instance.transient_data["node"].name():
-            continue
-
-        instance.data["active"] = True
+        instance.data["active"] = (
+            node.name() == instance.transient_data["node"].name()
+        )
 
     context = pyblish.api.Context()
     context.data["create_context"] = create_context


BigRoy:
Interesting that both of these PRs were created 19 minutes apart, but closing this in favor of https://github.com/ynput/ayon-nuke/pull/235

PR #235, opened 2026-04-08
When pressing the "Render On Farm" button on a write node it only renders that node

splidje:
## Changelog Description
When pressing the "Render On Farm" button on a write node it only renders that node rather than also rendering all other active write nodes locally.

### DIFF
--- a/client/ayon_nuke/api/utils.py
+++ b/client/ayon_nuke/api/utils.py
@@ -127,7 +127,7 @@ def _submit_render_on_farm(node) -> bool:
 
     # Ensure CreateInstance is enabled.
     for instance in create_context.instances:
-        instance.data["active"] = node.name() == instance.transient_data["node"].name()
+        instance.data["active"] = node == instance.transient_data["node"]
 
     context = pyblish.api.Context()
     context.data["create_context"] = create_context
--- a/client/ayon_nuke/api/utils.py
+++ b/client/ayon_nuke/api/utils.py
@@ -127,10 +127,7 @@ def _submit_render_on_farm(node) -> bool:
 
     # Ensure CreateInstance is enabled.
     for instance in create_context.instances:
-        if node.name() != instance.transient_data["node"].name():
-            continue
-
-        instance.data["active"] = True
+        instance.data["active"] = node.name() == instance.transient_data["node"].name()
 
     context = pyblish.api.Context()
     context.data["create_context"] = create_context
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/api/utils.py
```diff
@@ -127,10 +127,7 @@ def _submit_render_on_farm(node) -> bool:
 
     # Ensure CreateInstance is enabled.
     for instance in create_context.instances:
-        if node.name() != instance.transient_data["node"].name():
-            continue
-
-        instance.data["active"] = True
+        instance.data["active"] = node.name() == instance.transient_data["node"].name()
```
BigRoy: Can't we technically do a straight up comparison, or doesn't Nuke support that? Seems more reliable then matching by name (since technically node names are only unique under their parent groups, right?)

```suggestion
        instance.data["active"] = node == instance.transient_data["node"]
```

BigRoy: ~~Did a quick test - that doesn't seem to be an option.~~
~~Then let's do:~~


```suggestion
        instance_node = instance.transient_data["node"]
        instance.data["active"] = node.fullName() == instance_node .fullName()
```

Actually - straight up node comparison does work 🤔

BigRoy (approved):

client/ayon_nuke/api/utils.py
```diff
@@ -127,10 +127,7 @@ def _submit_render_on_farm(node) -> bool:
 
     # Ensure CreateInstance is enabled.
     for instance in create_context.instances:
-        if node.name() != instance.transient_data["node"].name():
-            continue
-
-        instance.data["active"] = True
+        instance.data["active"] = node == instance.transient_data["node"]
```
BigRoy:
```suggestion
        instance.data["active"] = node is instance.transient_data["node"]
```

moonyuet (commented):
Tested successfully in Nuke speaking of pressing the button in prerender and render product

Image

@BigRoy  @jakubjezek001 I assume pressing this button would publishing the prerender and render instance(aka sending to publish job to let ayon register the review) but not just the farm render submission. Is there any reason behind not integrating into the publish job? 
Image

This is not working for image instance of which also allows `Render On Farm` options as the relevant instance attributes enum

OlivierOsotimehin:
hello @moonyuet we would like to use this feature to quickly send a render or prerender to the farm without publishing so that we can then check and if good publish with using existing frame. 
The current workflow make the farm render & publish action tighly linked. This enable to decouple the farm rendering process and the publishing / web review process

moonyuet (approved):
The image instance issue appears to be another issue which is not related to this PR. We can fix in a separate PR instead.

PR #237, opened 2026-04-08
Load Clip: Allow setting frame range from version data if frame start or frame end is 0

BigRoy:
## Changelog Description

Load Clip: Allow setting frame range from version data if frame start or frame end is 0

...

## Testing notes:
1. Publish media (e.g. a video file) with a frame range version data like 0-100
2. When loading this clip the frame range should be set to 0-100 and go through the `_set_range_to_node` logic


moonyuet (approved):
I was using the same scene from https://github.com/ynput/ayon-nuke/pull/238 to test this PR.
tested successfully in Nuke 16 with the clip loader. 
EDIT: I got confused by these two PRs. Will be also approving the other one.
Image

PR #238, opened 2026-04-08
Load Clip: Set filepath with `fromUserText` to ensure video without frame range metadata on version sets frame range correctly

BigRoy:
## Changelog Description

Load Clip: Set filepath with `fromUserText` instead of first settng with `setValue` so that video files get their frame range set correctly instead of 1-1 when no frame range data exists on the version attributes.

It seems that loading e.g. a `.mov` publish that has no frame range metadata would previously not set the frame range correctly.
The real diffference is just using `fromUserText` on line 175 - the rest really is cosmetic changes that could be cleaned up along with that.

## Testing notes:

1. Load Clip should behave as you'd expect in all scenarios with regards to the frame ranges.

### DIFF
--- a/client/ayon_nuke/plugins/load/load_clip.py
+++ b/client/ayon_nuke/plugins/load/load_clip.py
@@ -182,9 +182,7 @@ class LoadClip(plugin.NukeLoader):
                     repre_entity
                 )
             if first is not None and last is not None:
-                self._set_range_to_node(
-                    read_node, first, last
-                )
+                self._set_range_to_node(read_node, first, last)
 
             if start_at_workfile:
                 self._start_at_workfile_frame(read_node, slate_frames)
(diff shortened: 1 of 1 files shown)


BigRoy:
PR also contains the change made in https://github.com/ynput/ayon-nuke/pull/237

moonyuet (approved):
Did simple test with some testing video from traypublisher in mp4,
tested successfully in Nuke 16 with the clip loader.

Image

moonyuet (approved):
I was using the same scene from https://github.com/ynput/ayon-nuke/pull/238 to test this PR.
tested successfully in Nuke 16 with the clip loader. 
EDIT: I got confused by two PRs (#237, #238) and approved twice. Will be also approving the other one.
Image

PR #242, opened 2026-04-11
enforce rgb as channels for intermediate review export mov

### LINKED ISSUE #208: YN-0549: nuke intermediate baking RGBA creates black thumbnail

The suspicion is that all channels will also bake into prores4444 mov more channels because that might be supported. And then ffmpeg actually convert it to png intermediate thumbnail file with alpha backed into it. So as you had mentioned only RGB channel setting is supported. 

Testing with develop branch of core addon
1. Create Render Instance(e.g. RenderTestAlpha) and exposes channels in `ayon+settings://nuke/create/CreateWriteRender/exposed_knobs` change the channel to `all` or `rgba`
2. Publish
3. Check the publish of the render instance e.g. RenderTestAlpha in the server
4. You would find the black thumbnail created for that render instance

Testing with the branch of core addon
1. Repeat the step 1-4 above
2.You would find the thumbnail created for that render instance

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1479127436712218847)

_This issue was automatically created from AYON ticket [YN-0549]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=07a88f60196311f1a3cfc14c8b06e5a0)

moonyuet:
## Changelog Description
This PR is to enforce the rgba as channels for the intermediate review export. By this, this would ensure the image won't be exported with black thumbnail as they are with the channels other than rgb.

Resolve https://github.com/ynput/ayon-nuke/issues/208

n/a

## Testing notes:
1. `ayon+settings://nuke/create/CreateWriteRender/knob_overrides` or similar setting for the write node creators to add the knob overridden value
2. Create Prerender or use any write node creator product type to create instance
3. Add the knob channels to display the channel data
4. Publish
5. Should show up the thumbnail as reviewables in the server

### DIFF
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1411,7 +1411,11 @@ class ExporterReviewMov(ExporterReview):
 
         write_node["file"].setValue(str(self.path))
         write_node["file_type"].setValue(str(self.ext))
-        write_node["channels"].setValue(str(self.color_channels))
+        skip_alpha_channel = kwargs.get("skip_alpha_channel", False)
+        if skip_alpha_channel:
+            write_node["channels"].setValue("rgb")
+        else:
+            write_node["channels"].setValue(str(self.color_channels))
 
         # Knobs `meta_codec` and `mov64_codec` are not available on centos.
         # TODO shouldn't this come from settings on outputs?
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -152,6 +152,10 @@ class IntermediateOutputModel(BaseSettingsModel):
         False,
         title="Publish"
     )
+    skip_alpha_channel: bool = SettingsField(
+        False,
+        title="Skip alpha channel for review intermediates"
+    )
     filter: BakingStreamFilterModel = SettingsField(
         title="Filter", default_factory=BakingStreamFilterModel)
     read_raw: bool = SettingsField(
@@ -376,6 +380,7 @@ DEFAULT_PUBLISH_PLUGIN_SETTINGS = {
             {
                 "name": "baking",
                 "publish": False,
+                "skip_alpha_channel": False,
                 "filter": {
                     "task_types": [],
                     "product_base_types": [],


client/ayon_nuke/api/plugin.py
```diff
@@ -1411,7 +1410,10 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
 
         write_node["file"].setValue(str(self.path))
         write_node["file_type"].setValue(str(self.ext))
-        write_node["channels"].setValue(str(self.color_channels))
+        # enforce this to rgb to be sure that we are not
+        # writing in other cololr channels than rgb and get
+        # the black thumbnail
+        write_node["channels"].setValue("rgb")
```
jakubjezek001: Yes that is the solution, but I can still imagine someone might need to be able to control it from settings. We could add actually some write node knob value overrides  so anybody could override some values, one of which is the channels. @BigRoy what do you thing? 

Or simplify it only to one single settings attribute `channels` with default value set to `rgb`. So anybody could at least change this to `rgba` if they need.

BigRoy: I'm mostly concerned as to WHY alpha would break things to begin with? Do we know why the thumbnail goes black to begin with? What makes that happen?

moonyuet: > I'm mostly concerned as to WHY alpha would break things to begin with? Do we know why the thumbnail goes black to begin with? What makes that happen?

It is not only alpha but the channels other than rgb. You can check the conversation in the linked issue.

client/ayon_nuke/api/plugin.py
```diff
@@ -1638,7 +1641,17 @@ def exposed_write_knobs(settings, plugin_name, instance_node):
     exposed_knobs = settings["nuke"]["create"][plugin_name].get(
         "exposed_knobs", []
     )
+    knob_overrides = settings["nuke"]["create"][plugin_name].get(
+        "knob_overrides", []
+    )
```
jakubjezek001: This will not be solution for the issue where if Render Write node is set to RGBA then intermediate files are also adding the Alpha into prores4444 files. I am assuming that best way would be to add toggle to Intermediate baking profile for `Pass Alpha channel too` or something like this. And set this by default to False. This way if some productions will need also Alpha in baked MOV (one client confirmed to me today they want alpha in them), that can be added. 

I agree with @BigRoy that we should figure out where why the Alpha in Intermediate baked MOV is killing thumbnail. If it is because thumbnail is created from the MOV file, then we need to find the process and constrain it to only use RGB from it.

jakubjezek001 (changes requested):
some of the latest changes in regards of passing Creators Knob overrides, they are adding complexity nobody will be using. 

- lets find out why thumbnail is not avoiding Alpha form baked MOV file
- lets add toggle for passing Alpha into intermediate baked MOV file on demand

moonyuet:
@jakubjezek001 @BigRoy I made the updated commit but I personally think that the `skip_alpha_channel` toggle is rather for the short-term fix and it is not also fully fixing the issue when we are talking about the encoding issue.
I am thinking we can have explicit thumbnail extractor, but we need to consider if it is worth for just thumbnail creation
Or we change the codec to different settings - I guess there should be reasons for using mov

moonyuet:
@jakubjezek001 @BigRoy I created the draft pull request in core addon: https://github.com/ynput/ayon-core/pull/1787 which might fix the issue of black thumbnail creation in Nuke.

moonyuet:
Close this as completed with https://github.com/ynput/ayon-core/pull/1787

PR #244, opened 2026-04-15
AYON dependency package slipped in front of the python path

splidje:
## Changelog Description
AYON dependency package slipped in front of the python path to override any modules packaged with nuke (e.g. opentimelineio)

### DIFF
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -2,6 +2,8 @@ import nuke
 
 import os
 import importlib
+import re
+import sys
 from collections import OrderedDict, defaultdict
 
 import pyblish.api
@@ -127,6 +129,21 @@ class NukeHost(
 
     def install(self):
         """Installing all requirements for Nuke host"""
+
+        # Ensure AYON dependencies
+        # override any modules packaged
+        # with Nuke. (e.g. opentimelineio)
+        for include_path in sys.path:
+            # TODO: this I know isn't a great way to do this.
+            # quite hacky. I asked for advice on Discord, but
+            # haven't seen a reply yet, but thought I'd push this
+            # so it's there to be seen.
+            if not re.search("/dependency_packages/.*/dependencies/?", include_path):
+                continue
+
+            sys.path.remove(include_path)
+            sys.path.insert(0, include_path)
+            break
 
         pyblish.api.register_host("nuke")
 


BigRoy (changes requested):
The intend is never for dependency package to **override** those from a DCC, but instead for our dependencies to complement the DCC. So the forced behavior here sounds incorrect.

My guess is we won't be merging this nor that we'll find the best solution here.

@iLLiCiTiT thoughts?

iLLiCiTiT:
Not sure how OTIO could slip into any DCC? It is runtime dependency -> only available for AYON launcher processes, never added to PYTHONPATH.

splidje:
I have a custom addon and I've added a particular version of otio as one of its dependencies, as I need a fix/feature it has.

I haven't yet seen whether overriding Nuke's version of otio is breaking Nuke in any way. Currently working ok.

splidje:
I guess I could instead put this code just inside my custom addon? As it's currently something only that custom addon needs?

I think I was just curious as to whether the general feature might be required by anyone else, but I get that overriding a DCC's packaged python modules feels a little dodgy.

iLLiCiTiT:
> I think I was just curious as to whether the general feature might be required by anyone else, but I get that overriding a DCC's packaged python modules feels a little dodgy.

You can't do that, especially with not pure python modules (like OTIO is), because they have to be build for the target python (and target OS is case of linux). That's why it is defined as runtime dependency, only for AYON launcher processes.

BigRoy:
Closing this PR - because it's not something we'll merge. But feel free to continue the discussion.

splidje:
If my addon running within Nuke needs a particular non-pure-python module, what's the correct way to do that?

iLLiCiTiT:
> If my addon running within Nuke needs a particular non-pure-python module, what's the correct way to do that?

No easy answer to that.

One option would be to install it using python and pip inside the DCC (if there is one).

If you want to avoid that, you can run AYON launcher process with some socket server where you'd process what you need to process using some communications logic. That is very hard and far from the standard usage of python module.

This is something that is not easy to tackle, is based on the usecase and knowledge of a developer.

splidje:
Yes, I have done that in the past with running a subprocess, but always nice to avoid it.

I've usually in the past built various versions of the python module to match whichever combinations of OS+python version+etc I need (i.e. depending on the various DCCs in use) then set things up to switch to the relevant one depending on the circumstances.

Is it fair to say that that sort of thing is in the realms of something like rez and the AYON dependency features are solely about pure python modules? i.e. should I look beyond AYON to achieve this particular thing?

BigRoy:
> Is it fair to say that that sort of thing is in the realms of something like rez and the AYON dependency features are solely about pure python modules? i.e. should I look beyond AYON to achieve this particular thing?

Currently this is beyond what AYON can do, but I wouldn't see it's off the table for AYON completely.

The concept has been in the past to see if we can ship on the fly environments per process, so that e.g. we can serve different dependencies depending on what version of nuke, etc. but the complexity hasn't been worth it so far. We've looked into Rez in the past but failed to get something we were satisfied with, and have recently played with the idea of using `uv` to construct runtime environments as you start an application. - But nothing that exists out of the box currently.

iLLiCiTiT:
It's not out of view but it has it's cost.

First of all, development time, and massive changes across all repositories. Secondly, we have to ship uv with ayon launcher, which might be challenging.

Then you must have internet connection to be able to use uv. Then you have to have more strict rules about application executable paths, because currently you can use `.bat` `.sh` etc, which won't be possible if we have to rely on being 100% be able to find the correct python executable. And few more things we'd not be able to support with that anymore.

PR #245, opened 2026-04-20
Fixing effect loader for macos resources 

jakubjezek001:
## Changelog Description
Fixes effect loader on macOS by properly resolving resource file paths using AYON Anatomy templates. This ensures effects reference resources with correct platform-specific paths, resolving issues with effect loading on macOS systems.

## Additional info
The loader now uses the Anatomy system to convert absolute paths into platform-agnostic template paths. When setting node attributes with file references, the code:
1. Detects file path knobs in the effect node structure
2. Retrieves the Anatomy configuration for the current project
3. Converts absolute paths to rootless template paths
4. Resolves the template with current root mappings to generate correct platform-specific paths

This approach ensures effects work correctly across different OS environments and project configurations.

## Testing notes:
1. Load an effect with external resources on macOS
2. Verify effect resources are correctly located and loaded
3. Test with multiple project root configurations
4. Confirm effects function identically across Windows, macOS, and Linux


client/ayon_nuke/plugins/load/load_effects.py
```diff
@@ -65,9 +65,8 @@ def connect_read_node(self, group_node, namespace, product_name):
     def _load_effects_to_group(
             self, context: dict, group_node: nuke.Node) -> str:
         """Load the json file and create nodes inside the group node"""
-
         file = self.filepath_from_context(context).replace("\\", "/")
-        with open(file, "r") as f:
+        with open(file) as f:
```
iLLiCiTiT:
```suggestion
        with open(file, "r") as f:
```

client/ayon_nuke/plugins/load/load_effects.py
```diff
@@ -103,6 +101,26 @@ def _create_nodes_order(self, nodes_order: dict):
                     self.log.warning(e)
                     continue
 
+                # check if `file` in knob name then format the path
+                # so it is multiplatform compatible
+                if k == "file":
+                    file_path =  v
+                    project_name = get_current_project_name()
+                    anatomy = Anatomy(project_name=project_name)
```
iLLiCiTiT, line 109: I think this should use the project from which we're loading. Mostly would be current project, but you can also load from library project. The project entity is stored in `context`, so you could just do 
```python
project_entity = context["project"]
anatomy = Anatomy(project_entity["name"], project_entity=project_entity)
```

Also could happen only once for the whole load logic in case the anatomy would be created multiple times.

iLLiCiTiT: Side note: Why is this actually happening? The loaded "files" can have loaded other files? That might be a mess.  Imagine having file that is e.g. ocio file defined with environment variables instead of anatomy roots, it would just collapse.

client/ayon_nuke/plugins/load/load_effects.py
```diff
@@ -103,6 +101,26 @@ def _create_nodes_order(self, nodes_order: dict):
...
+                    success, rootless_path = \
+                        anatomy.find_root_template_from_path(file_path)
+
+                    self.log.info(f"rootless_path: {rootless_path}")
+                    if success:
+                        abs_resources_path = StringTemplate.format_strict_template(
+                            rootless_path, _template_data
+                        )
+                        v = abs_resources_path
+                        self.log.info(f"File path: {abs_resources_path}")
```
iLLiCiTiT:
```suggestion
```

client/ayon_nuke/plugins/load/load_effects.py
```diff
@@ -103,6 +101,26 @@ def _create_nodes_order(self, nodes_order: dict):
                     self.log.warning(e)
                     continue
 
+                # check if `file` in knob name then format the path
+                # so it is multiplatform compatible
+                if k == "file":
+                    file_path =  v
+                    project_name = get_current_project_name()
```
iLLiCiTiT, line 108: 2 things:
- it is possible to load from other projects
- the Anatomy creation can be very heavy operation

Rather prepare the anatomy ahead in `_load_effects_to_group` and use project entity from `context`.
```python
project_entity = context["project"]
anatomy = Anatomy(
    project_entity["name"], project_entity=project_entity
)
```

client/ayon_nuke/plugins/load/load_effects.py
```diff
@@ -103,6 +101,27 @@ def _create_nodes_order(self, nodes_order: dict):
                     self.log.warning(e)
                     continue
 
+                # check if `file` in knob name then format the path
```
BigRoy, line 104: @jakubjezek001 please check your phrasing when writing whether a string matches something. The check is not whether **file** is IN the knob name, but whether it is IS the knob name. It's an important difference and I've seen quite a bit of where we did `if "text" in value:` where we should've been doing `if value == "text"` or alike. In this case just the comment is wrong, but I want to just emphasize that we should make sure what it's saying is correct.
```suggestion
                # If it is the "file" knob then format the path
```

PR #251, opened 2026-05-01
Load Backdrop: Avoid scene connections breaking to selected nodes when loading backdrop

BigRoy:
## Changelog Description

Fixes issue where the scene may be losing existing connections if loading backdrop. The easiest way to reproduce is to select all nodes in the graph, load the placeholder. Previously it'd essentially break your scene.

Reset selection before creating the NoOp to avoid the selection getting automatically connected to the input of the NoOp (default nuke behavior) + move the logic inside the maintained selection contextlib so that we do actually maintain the users selection as intended. Also, remove some redundant `reset_selection` calls before explicitly deleting a node (selection should not affect it.)

## Testing notes:

1. Load Placeholder should work.

### DIFF
--- a/client/ayon_nuke/plugins/workfile_build/load_placeholder.py
+++ b/client/ayon_nuke/plugins/workfile_build/load_placeholder.py
@@ -358,7 +358,7 @@ class NukePlaceholderLoadPlugin(NukePlaceholderPlugin, PlaceholderLoadMixin):
 
         # Workaround: There are some cases where only the second call to `.dependent()` seems to
         #  start returning the input dependencies directly after a scene open (reproduced in Nuke 14.1)
-        #  so we just enforce and extra call to be sure this works as intended
+        #  so we just enforce an extra call to be sure this works as intended
         placeholder_node.dependent()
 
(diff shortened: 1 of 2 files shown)


moonyuet (approved):
Works with testing with import backdrop and load placeholder in Nuke 16.0

client/ayon_nuke/plugins/workfile_build/load_placeholder.py
```diff
@@ -355,6 +355,12 @@ def _set_loaded_connections(self, placeholder):
         input_node, output_node = get_group_io_nodes(
             placeholder.data["last_loaded"]
         )
+
+        # Workaround: There are some cases where only the second call to `.dependent()` seems to
+        #  start returning the input dependencies directly after a scene open (reproduced in Nuke 14.1)
+        #  so we just enforce and extra call to be sure this works as intended
```
BigRoy:
```suggestion
        #  so we just enforce an extra call to be sure this works as intended
```

jakubjezek001 (approved):

PR #254, opened 2026-05-11
Version handles calculated based on context handles

splidje:
## Changelog Description
Version handles calculated based on context handles. Handles set for plate product type as well as render.


client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -172,9 +172,9 @@ def _set_additional_instance_data(
         color_channels = write_node["channels"].value()
 
         # get frame range data
-        handle_start = instance.context.data["handleStart"]
-        handle_end = instance.context.data["handleEnd"]
         first_frame, last_frame = self._get_frame_range_data(instance)
+        handle_start = max(0, root_first_frame + context_handle_start - first_frame)
+        handle_end = max(0, last_frame - (root_last_frame - context_handle_end))
```
rdelillo: I'm not sure I understand where `root_first_frame` and `context_handle_start` come from.
Also do you  mind to detail a bit more on why you need this new logic replacing `instance.context_data` please ?

rdelillo (changes requested):

splidje:
I've now realised this is kind of _utterly_ flawed!

The only thing I really need to hit - based on how the load currently works - is making sure the version start handle is the same as the shot start handle, as the load logic takes the timewarp data to be starting from script start + version start handle... then the algorithm just runs through the list one by one irrespective of any end handles.

It kind of feels like it might make sense to set the version end handle in some kind of graceful way. i.e. set it to a number which consumes all the frames at the end of the version's frame range which end up falling outside of the shot's frame range (sans handles) having applied the timewarp.

Further to that, it feels that the same could be done with the version's start handle (but currently because of the load logic, it must be the same as the shot's start handle).

Feels like the version's start/end handle should be based on which frames at the beginning and the end don't make it inside the shot's frame range after applying the timewarp.

rdelillo:
@splidje based on your last comment ? should this PR still be reviewed ?
I'm wondering if we shouldn't create an issue for your use-case and discuss its implementation there first.

PR #256, opened 2026-05-19
Added checkbox to write node `requires_gpu`

splidje:
## Changelog Description
Added checkbox to write node `requires_gpu` as a shortcut to setting the OpenCue publish attribute, so "Render On Farm" button picks it up.

Works bi-directionally - i.e. changing the publish attribute updates the checkbox and vice versa.


BigRoy (changes requested):
Logic is essentially specific to some OpenCue integration - and hence I'd argue it should actually live there. 

Unless this has usage outside of OpenCue and we can change the code accordingly to generalize, but I don't think that's the case.

splidje:
@BigRoy , is there a way from the ayon_opencue addon to hook into the ayon_nuke write node and add this boolean knob?

BigRoy:
> @BigRoy , is there a way from the ayon_opencue addon to hook into the ayon_nuke write node and add this boolean knob?

I don't think there is by current design no. But is it important to be its own knob on the node, instead of having it available in the publisher UI. Because _that_ is possible from other plug-ins, yes.

splidje:
Sorry for late reply.

I have my artists using the "Render On Farm" button on the write node, in order to try renders as many times as they want before publishing.

Showing them the workflow of:
1. AYON -> Publish
2. Publish tab
3. Select a write node
4. Change "Requires GPU"
5. Close window so it saves it
6. Click "Render On Farm"

Was clearly quite unfriendly, so I told them I'd also add a knob on the write node to control that setting.

Which then becomes:
1. Change "Requires GPU" knob
2. Click "Render On Farm"

BigRoy:
It's too specific to OpenCue and not generic enough currently.

Will need to be adapted - or we'll need to close the PR.

splidje:
Looking again at this today, I reckon I can move the logic to the opencue add-on.
Either have it register an on create callback or perhaps a more AYON-ey create instance created callback?

splidje:
There: https://github.com/splidje/ayon-opencue/commit/af8306c14662f67cb1a31308a444990659ae6716

This has been moved to the opencue repo now.

Issue #262, opened 2026-06-10
CollectWrite should not create a representation if render target is farm


sebastianbrandhuber:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
The collect_writes plugin currently adds a representation even if render target is set to farm.

## Expected Behavior
If instance.data["farm"] is True, no representation should be added since processing should happen on farm.

### Steps To Reproduce:

1. Create a write instance
2. Publish with render target set to "use existing frames - farm"

### Version
ayon-core 1.9.4, ayon-nuke 0.4.10

### What platform were you running when you found the bug?
Windows 11

Link to the discussion is [here](https://community.ynput.io/t/possible-missing-farm-check-in-extractcolorspacetranscode/3214/4)

I will create a PR for this and link it to the issue.

### Relevant log output:

```shell

```

jakubjezek001:
closed by https://github.com/ynput/ayon-nuke/pull/263

PR #263, opened 2026-06-10
collect_writes - skip adding representation for instance render target "farm"

sebastianbrandhuber:
## Changelog Description
During collection of write nodes, if instance.data["farm"] is True, no representation should be added since processing should happen on farm.

Link to issue [here](https://github.com/ynput/ayon-nuke/issues/262)

## Testing notes:
1. create a nuke write instance
2. switch render target to "use existing frames - farm"
3. publish -> processing should happen on farm only

### DIFF
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -48,11 +48,11 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         colorspace = napi.get_colorspace_from_node(write_node)
 
         if render_target == "frames":
-            self._set_existing_files_data(instance, colorspace)
+            self._set_existing_files_data(instance, colorspace, render_target)
 
         elif render_target == "frames_farm":
             collected_frames = self._set_existing_files_data(
-                instance, colorspace)
+                instance, colorspace, render_target)
 
             self._set_expected_files(instance, collected_frames)
 
@@ -64,17 +64,22 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         # set additional instance data
         self._set_additional_instance_data(instance, render_target, colorspace)
 
-    def _set_existing_files_data(self, instance, colorspace):
+    def _set_existing_files_data(self, instance, colorspace, render_target):
         """Set existing files data to instance data.
 
         Args:
             instance (pyblish.api.Instance): pyblish instance
             colorspace (str): colorspace
+            render_target (str): instance render target
 
         Returns:
             list: collected frames
         """
         collected_frames = self._get_collected_frames(instance)
+
+        # Skip adding representation if processed on farm
+        if render_target == "frames_farm":
+            return collected_frames
 
         representation = self._get_existing_frames_representation(
             instance, collected_frames
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -48,11 +48,10 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         colorspace = napi.get_colorspace_from_node(write_node)
 
         if render_target == "frames":
-            self._set_existing_files_data(instance, colorspace, render_target)
+            self._set_existing_files_data(instance, colorspace)
 
         elif render_target == "frames_farm":
-            collected_frames = self._set_existing_files_data(
-                instance, colorspace, render_target)
+            collected_frames = self._get_collected_frames(instance)
 
             self._set_expected_files(instance, collected_frames)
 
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -48,11 +48,11 @@ def process(self, instance):
         colorspace = napi.get_colorspace_from_node(write_node)
 
         if render_target == "frames":
-            self._set_existing_files_data(instance, colorspace)
+            self._set_existing_files_data(instance, colorspace, render_target)
 
         elif render_target == "frames_farm":
             collected_frames = self._set_existing_files_data(
-                instance, colorspace)
+                instance, colorspace, render_target)
```
BigRoy: I think the solution here is to just call `collected_frames = self._get_collected_frames(instance)` here instead.

BigRoy (changes requested):

client/ayon_nuke/plugins/publish/collect_writes.py
sebastianbrandhuber: yep, makes sense,
I fixed it; still works as expected.

BigRoy (approved):
LGTM - @rdelillo @jakubjezek001 can you confirm that this actually makes sense. Essentially the goal was to make sure `instance.data["representations"]` does not get populated if we're targeting the farm.

rdelillo:
@sebastianbrandhuber just to be sure, would you mind to elaborate a bit more on the actual issue you're facing ?
Just checked, even if a representation get added to the instance, it does not integrate locally when `frames_farm` is selected.
Is the data collection failing ? Do you see that more an as an optimization ? 

image

I personally don't mind the change but just wondering if  collecting a representation here might not be beneficial for some validation purpose @jakubjezek001.

sebastianbrandhuber:
@rdelillo sure,
I had a brief discussion with @BigRoy and @iLLiCiTiT in the community forum about it:
https://community.ynput.io/t/possible-missing-farm-check-in-extractcolorspacetranscode/3214/4

Basically, the fix tries to ensure plugins like ExtractColorspaceTranscode are not being processed during publish at all if the instance is not being processed locally - especially if those plugins do not check if the process should happen on farm or not.

I just added a PR on ayon-core regarding the issue for the ExtractColorspaceTranscode plugin,
but other plugins might face this issue aswell.

> I personally don't mind the change but just wondering if collecting a representation here might not be beneficial for some validation purpose

Valid point. I'd be interested to hear other people's thoughts on this.

rdelillo:
> @rdelillo sure, I had a brief discussion with @BigRoy and @iLLiCiTiT in the community forum about it: https://community.ynput.io/t/possible-missing-farm-check-in-extractcolorspacetranscode/3214/4
> 
> Basically, the fix tries to ensure plugins like ExtractColorspaceTranscode are not being processed during publish at all if the instance is not being processed locally - especially if those plugins do not check if the process should happen on farm or not.
> 
> I just added a PR on ayon-core regarding the issue for the ExtractColorspaceTranscode plugin, but other plugins might face this issue aswell.
> 
> > I personally don't mind the change but just wondering if collecting a representation here might not be beneficial for some validation purpose
> 
> Valid point. I'd be interested to hear other people's thoughts on this.

Thanks for the link. I missed that. I'm gonna go ahead and approve this.
Unless @jakubjezek001 has an objection, I'd say this is good to go. Thanks for the fix.

rdelillo (approved):

iLLiCiTiT:
Validator `ValidateRenderedFrames` is using `"representations"`, not sure if it should validate the range?

BigRoy:
> Validator `ValidateRenderedFrames` is using `"representations"`, not sure if it should validate the range?

Seems like a slightly odd validator to begin with. @jakubjezek001 do you know more about the intent and whether this would be a problem? or do we remove that validator, or should we actually validate something else?

moonyuet:
> > Validator `ValidateRenderedFrames` is using `"representations"`, not sure if it should validate the range?
> 
> Seems like a slightly odd validator to begin with. @jakubjezek001 do you know more about the intent and whether this would be a problem? or do we remove that validator, or should we actually validate something else?

I agree that it is a bit odd validator, but it does check if there is any expected files already existed before the farm publish.
Shall we open another issue for this and move the convo over there?

moonyuet (approved):
Tested successfully in Nuke 

[publish-report-260616-21-24.json](https://github.com/user-attachments/files/29003176/publish-report-260616-21-24.json)

BigRoy:
> I agree that it is a bit odd validator, but it does check if there is any expected files already existed before the farm publish.
> Shall we open another issue for this and move the convo over there?

If it gets broken/disabled by this PR because we don't have the data in the right place and suddenly things are not validated - I think we should make sure it doesn't break. :) So does the validation still do what it should? if so, all good to merge. Did we introduce a regression - then we should fix it.

moonyuet:
> > I agree that it is a bit odd validator, but it does check if there is any expected files already existed before the farm publish.
> > Shall we open another issue for this and move the convo over there?
> 
> If it gets broken/disabled by this PR because we don't have the data in the right place and suddenly things are not validated - I think we should make sure it doesn't break. :) So does the validation still do what it should? if so, all good to merge. Did we introduce a regression - then we should fix it.

Double check again. It doesn't break the validation.

PR #266, opened 2026-06-10
Add context reset prompt on taskChange

### LINKED ISSUE #265: Add nuke context reset prompt on taskChange
**Is your feature request related to a problem? Please describe.**
When saving a workfile into a different context (e.g., a different task), only the menu label is currently updated.

**Describe the solution you'd like**
Similar to ayon-houdini, we should display a pop-up dialog that allows users to interactively choose whether, and what, should be adapted to the new context. An example of this approach can be seen here:

https://github.com/ynput/ayon-houdini/blob/87c32b5ab6c967f34c3b6eb1b0349b23caa2b55b/client/ayon_houdini/api/lib.py#L1504

**Describe alternatives you've considered**
While the Houdini approach could theoretically be implemented in a similar way for Nuke, I am also considering an alternative solution that replaces existing instances instead of updating their attributes.

**Additional context**
A community post about this can be found here: 
https://community.ynput.io/t/nuke-context-switch-prompt-context-reset/3205
I will create a basic PR for testing and discussion using the Houdini-style approach.

sebastianbrandhuber:
## Changelog Description
This PR implements a pop-up dialog that allows users to interactively choose whether, and what, should be adapted to the new context before a context change callback - very similar to its counterpart in ayon-houdini.

Additional information can be found here in the issue itself:
Fix https://github.com/ynput/ayon-nuke/issues/265

## Testing notes:
1. Start nuke, save a scene with the workfiles tool.
2. (Optional) Create a write instance
3. Using the workfiles tools, save this existing scene into a different context
4. A dialog shows up and lets the user choose what should be adapted to the new context.

## Current limitations and unintended bahaviour:
- The product name of the instance does not get updated automatically - visible in the publish dialog.
- The node name of the instance node does not get updated
- The instance node's group children are not being renamed aswell.


client/ayon_nuke/api/lib.py
```diff
@@ -2560,6 +2567,86 @@ def _launch_workfile_app():
...
+
+    create_context = CreateContext(host, reset=True)
+
+    for instance in create_context.instances:
+        instance_folder_path = instance.get("folderPath")
+        if instance_folder_path and instance_folder_path != folder_path:
+            instance["folderPath"] = folder_path
+        instance_task = instance.get("task")
+        if instance_task and instance_task != task:
+            instance["task"] = task
```
sebastianbrandhuber: Another alternative approach for updating the instances that come to my mind 
could be to replace the instances (=their nodes) completely and delete the old ones.

Here is a quick sketch of the idea:

```
for instance in create_context.instances:
    if (
        instance.get("folderPath") in {None, folder_path}
        and instance.get("task") in {None, task}
    ):
        continue

    if instance.creator_identifier == "workfile":
        continue

    old_node = instance.transient_data["node"]
    if not old_node:
        continue
    old_node.knob("name").setValue(f"{old_node.name()}__old")

    new_instance = create_context.create(
        creator_identifier=instance.creator_identifier,
        variant=instance.get("variant", ""),
        pre_create_data={"use_selection": False},
    )

    new_node = new_instance.transient_data["node"]

    if old_node.inputs():
        new_node.setInput(0, old_node.input(0))

    new_node.setXYpos(old_node.xpos(), old_node.ypos())

    nuke.delete(old_node)
```

This would solve any mentioned issues in the PR,
additionally any preset values or naming conventions via project or site settings would be applied.
Still feels a bit hacky, though.

BigRoy: I wouldn't go that far. I do think that perhaps we may want to get the creator of each instance, and update product name with the result of `creator.get_product_name`? 🤔 And then if it differs update the product name and save it. If node names should then update along, it's really up to the creator to implement so in `update_instances` when relevant attribute, e.g. the product name get updated.

But I would definitely not go as far as deleting and recreating - just too many potential intentional customizations one may have applied in the scene that could be lost but should be maintained.

BigRoy (commented):
Nice - very nice! Code looks very close to similar functionality in e.g. Maya, Houdini and Fusion integration - and I like such consistency!

client/ayon_nuke/api/lib.py
sebastianbrandhuber: Right, what about something like this:

```
def update_content_on_context_change():
    """Update creator instances when the current folder/task changes."""

    host = registered_host()
    context = host.get_current_context()

    current_folder_path = context["folder_path"]
    current_task = context["task_name"]

    create_context = CreateContext(host, reset=True)

    project_entity = create_context.get_current_project_entity()
    folder_entity = create_context.get_current_folder_entity()
    task_entity = create_context.get_current_task_entity()

    project_name = project_entity["name"]
    host_name = create_context.host_name

    _changed = False

    for instance in create_context.instances:
        if instance.creator_identifier == "workfile":
            continue

        if (
            instance.get("folderPath") == current_folder_path
            and instance.get("task") == current_task
        ):
            continue

        creator = create_context.creators.get(
            instance.creator_identifier
        )
        if creator is None:
            continue

        product_name = creator.get_product_name(
            project_name=project_name,
            project_entity=project_entity,
            folder_entity=folder_entity,
            task_entity=task_entity,
            variant=instance.get("variant"),
            host_name=host_name,
        )

        instance["folderPath"] = current_folder_path
        instance["task"] = current_task
        instance["productName"] = product_name

        _changed = True

    if _changed:
        create_context.save_changes()
```

This still does not include anything a fresh write node would get:
- get knob settings for write node
- rename the actual write node inside the group
- create directory for new write path
- adding render farm submission button based on settings

But as you said,
this could be implemented in the creator itself in another PR.

sebastianbrandhuber: Also, might be a different topic, but I just noticed that
the ruff.toml in ayon-nuke is less permissive than the config file in ayon-core or ayon-houdini.
Any reason why?
The biggest difference would be in select:
`select = ["E", "F", "W"]` vs `select = ["E4", "E7", "E9", "F"]`
If not a topic for this issue, let me know where this belongs!

BigRoy: @jakubjezek001 @rdelillo any opinions here?

client/ayon_nuke/api/lib.py
```diff
@@ -2560,6 +2567,120 @@ def _launch_workfile_app():
...
 
 
+def update_content_on_context_change():
+    """Update creator instances when the current folder/task changes."""
+
+    host = registered_host()
+    context = host.get_current_context()
+
+    current_folder_path = context["folder_path"]
+    current_task = context["task_name"]
```
BigRoy: I think we can skip these - and just rely on the `folder_entity["path"]` and `task_entity["name"]` below from the entities we get from the `CreateContext` (which initializes to the same context of the registered host).

client/ayon_nuke/api/lib.py
```diff
@@ -2560,6 +2567,117 @@ def _launch_workfile_app():
...
+    create_context = CreateContext(host, reset=True)
+
+    project_entity = create_context.get_current_project_entity()
+    folder_entity = create_context.get_current_folder_entity()
+    task_entity = create_context.get_current_task_entity()
+
+    current_folder_path = folder_entity.get("path")
+    current_task = task_entity.get("name")
+
+    project_name = project_entity.get("name")
```
BigRoy: These should really exist - if they do not, we have bigger problems and should error 🗡️ 
```suggestion
    current_folder_path = folder_entity["path"]
    current_task = task_entity["name"]

    project_name = project_entity["name"]
```

client/ayon_nuke/api/lib.py
```diff
@@ -2560,6 +2567,117 @@ def _launch_workfile_app():
...
+            instance.get("folderPath") == current_folder_path
+            and instance.get("task") == current_task
+        ):
+            continue
+
+        creator = create_context.creators.get(
+            instance.creator_identifier
+        )
+        if creator is None:
+            continue
```
iLLiCiTiT:
```suggestion
        creator = create_context.creators[instance.creator_identifier]
```

client/ayon_nuke/api/lib.py
BigRoy: > The biggest difference would be in select:
> `select = ["E", "F", "W"]` vs `select = ["E4", "E7", "E9", "F"]`
> If not a topic for this issue, let me know where this belongs!

We prefer stricter, so Nuke is 'ahead' here. The others just haven't followed suit yet.

client/ayon_nuke/api/pipeline.py
```diff
@@ -405,6 +411,28 @@ def change_context_label():
...
+    _about_to_save = False
+
+
+def before_workfile_save(event):
+    global _about_to_save
+    _about_to_save = True
+
+
+def on_task_changed():
+    global _about_to_save
```
iLLiCiTiT: Should not be needed if you're just checking the value.
```suggestion
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -138,8 +142,10 @@ def install(self):
         register_workfile_build_plugin_path(WORKFILE_BUILD_PATH)
 
         # Register AYON event for workfiles loading.
+        register_event_callback("workfile.save.before", before_workfile_save)
+        register_event_callback("workfile.save.after", after_workfile_save)
         register_event_callback("workio.open_file", check_inventory_versions)
-        register_event_callback("taskChanged", change_context_label)
+        register_event_callback("taskChanged", on_task_changed)
```
iLLiCiTiT: BTW host is responsible for context change, and default implementation does trigger `_after_context_change` method on host, which is probably where you should handle this instead of callback?

client/ayon_nuke/api/pipeline.py
BigRoy: It does make it rather explicit that it's taking that value though, perhaps?

client/ayon_nuke/api/pipeline.py
BigRoy: Oh, I didn't know about that one - nice!

client/ayon_nuke/api/lib.py
sebastianbrandhuber: Just to double check, by "stricter" you mean follow more rules?
Then I think Nuke is actually behind, right? Linting in the ruff.toml in ayon-core looks stricter than in ayon-nuke to me.

client/ayon_nuke/api/pipeline.py
sebastianbrandhuber: Found it, thanks.
Will implement it in NukeHost and remove from the callbacks.

client/ayon_nuke/api/lib.py
BigRoy: Sorry, I hadn't looked the configs myself, just went based on your comment: 

> the ruff.toml in ayon-nuke is less permissive than the config file in ayon-core or ayon-houdini.

Less permissive sounded like "more strict". But I see now that Nuke is the one on `["E4", "E7", "E9", "F"]` so yes - that should be made stricter. In a dedicated PR perhaps.

sebastianbrandhuber: Ah yeah sorry, I meant **more permissive** for ayon-nuke.

client/ayon_nuke/api/pipeline.py
```diff
@@ -162,6 +168,25 @@ def update_context_data(self, data, changes):
...
+            context_change_data (ContextChangeData): Object with information
+                about context change.
+
+        """
+        if not nuke.GUI:
+            return
+
+        change_context_label()
+
+        if _about_to_save:
```
iLLiCiTiT: Also pointing out that there is also `_before_workfile_save` and `_after_workfile_save` on `IWorkfileHost` so you can store the `_about_to_save` on the host itself too.

BigRoy:
@moonyuet could you give this a thorough stress test? :)

client/ayon_nuke/api/pipeline.py
BigRoy: @sebastianbrandhuber I assume this is a change still to do?

sebastianbrandhuber: Yes, will implement this next week when back at work 👍

moonyuet (changes requested):
The context change is successfully when switching folder and task. but failed to save the workfile.

https://github.com/user-attachments/assets/ea3e769e-19fa-4596-908f-60ff38ae613b

Error log:
```
  File "<string>", line 9, in <module>
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 287, in __init__
    self.reset(discover_publish_plugins)
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 580, in reset
    self.reset_instances()
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 1591, in reset_instances
    raise CreatorsCollectionFailed(failed_info)
ayon_core.pipeline.create.exceptions.CreatorsCollectionFailed: Failed to collect instances
```

client/ayon_nuke/api/lib.py
```diff
@@ -2568,6 +2575,113 @@ def _launch_workfile_app():
     host_tools.show_workfiles(parent=None, on_top=True)
 
 
+def update_content_on_context_change():
+    """Update creator instances when the current folder/task changes."""
+    host = registered_host()
+    create_context = CreateContext(host, reset=True)
```
moonyuet: @BigRoy @iLLiCiTiT  @sebastianbrandhuber Should we make sure the `publish_plugins` not being discovered as if we did in Maya?

BigRoy: So like this? (`reset` defaults to `True` so we can skip it?) 
```suggestion
    create_context = CreateContext(host, discover_publish_plugins=False)
```

Makes sense, would make it a bit faster. Whether it'll have side effects we'll need to double check. 

Can you make sure to test @moonyuet. Set a publish attribute in the publisher UI, save. Then save to another context, accept the prompt. Did the plug-in publish attribute override survive? or did it get lost because now we saved the Create Context without knowing about the plug-ins to begin with? 🤔

BigRoy:
> ```
>   File "<string>", line 9, in <module>
>   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 287, in __init__
>     self.reset(discover_publish_plugins)
>   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 580, in reset
>     self.reset_instances()
>   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 1591, in reset_instances
>     raise CreatorsCollectionFailed(failed_info)
> ayon_core.pipeline.create.exceptions.CreatorsCollectionFailed: Failed to collect instances
> ```

That sounds quite weird. Does the publisher UI refresh without errors before the save change? And does it still open publisher UI fine after?

It's weird it fails on collecting instances, but doesn't in stack trace tell us what it actually errors on.

moonyuet:
> > ```
> > Traceback (most recent call last):
> >   File "<string>", line 9, in <module>
> >   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 287, in __init__
> >     self.reset(discover_publish_plugins)
> >   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 580, in reset
> >     self.reset_instances()
> >   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\core_1.9.6+dev-ayon-ui-qt-launcher\ayon_core\pipeline\create\context.py", line 1591, in reset_instances
> >     raise CreatorsCollectionFailed(failed_info)
> > ayon_core.pipeline.create.exceptions.CreatorsCollectionFailed: Failed to collect instances
> > ```
> 
> That sounds quite weird. Does the publisher UI refresh without errors before the save change? And does it still open publisher UI fine after?
> 
> It's weird it fails on collecting instances, but doesn't in stack trace tell us what it actually errors on.

It does have error from ayon-deadline-cloud, I will try to disabled that plugins and test if it is related.

moonyuet (approved):
Tested successfully in Nuke 16.0

client/ayon_nuke/api/lib.py
moonyuet: @BigRoy I tested and it works now. It is due to some issue from ayon-deadline-cloud, which is not related to Nuke integration. I also tested with the code changes above. It is working too for what you have asked.

BigRoy (changes requested):
All looks good - just pending the change requested [here](https://github.com/ynput/ayon-nuke/pull/266#discussion_r3441027999) next week.

jakubjezek001 (approved):
works perfect and I'd be very happy to merge once requested changes from @BigRoy would be satisfied.

client/ayon_nuke/api/pipeline.py
sebastianbrandhuber: @BigRoy @iLLiCiTiT 
We could implement `about_to_save` directly as a property in `IWorkfileHost` within ayon-core so that other addons can benefit from it later - or leave it in the dcc addons and override the methods. What do you think?

BigRoy: For now create it in this host addon specifically for Nuke - then create an issue on core to continue the discussion. It'd need to be done here in backwards compatible way anyway so let's make sure we can move forward.

BigRoy: @sebastianbrandhuber this is still pending anything from our end?

sebastianbrandhuber: @BigRoy I just pushed the commit for review

client/ayon_nuke/api/pipeline.py
```diff
@@ -163,6 +168,60 @@ def update_context_data(self, data, changes):
...
+
+        Args:
+            save_workfile_context (SaveWorkfileContext): Workfile path with
+                target folder and task context.
+
+        """
+        super()._before_workfile_save(save_workfile_context)
+        self.about_to_save = True
+
+
```
BigRoy:
```suggestion
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -163,6 +168,60 @@ def update_context_data(self, data, changes):
...
+        Args:
+            save_workfile_context (SaveWorkfileContext): Workfile path with
+                target folder and task context.
+
+        """
+        super()._after_workfile_save(save_workfile_context)
+        self.about_to_save = False
+
+
```
BigRoy:
```suggestion
```

BigRoy (approved):
Aside of cosmetics (~~see suggestions~~ committed them) code looks good to me.

@moonyuet can you do a new test run?

client/ayon_nuke/api/pipeline.py
BigRoy: Looking great!

client/ayon_nuke/api/lib.py
```diff
@@ -2568,6 +2575,113 @@ def _launch_workfile_app():
...
+    """Update creator instances when the current folder/task changes."""
+    host = registered_host()
+    create_context = CreateContext(host, discover_publish_plugins=False)
+
+    project_entity = create_context.get_current_project_entity()
+    folder_entity = create_context.get_current_folder_entity()
+    task_entity = create_context.get_current_task_entity()
+
+    current_folder_path = folder_entity["path"]
+    current_task = task_entity["name"]
```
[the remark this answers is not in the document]
BigRoy: This is never the case in a reasonable setting.

client/ayon_nuke/api/lib.py
```diff
@@ -2568,6 +2575,113 @@ def _launch_workfile_app():
...
+    if options["resolution"]:
+        WorkfileSettings().reset_resolution()
+    if options["frame_range_fps"]:
+        WorkfileSettings().reset_frame_range_handles()
+    if options["colorspace"]:
+        WorkfileSettings().set_colorspace()
+    if options["instances"]:
+        update_content_on_context_change()
+
+    dialog.deleteLater()
```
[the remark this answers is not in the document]
BigRoy: I think this is useful fix actually - especially the re-use of the `WorkfileSettings()`.

client/ayon_nuke/api/lib.py
```diff
@@ -2568,6 +2575,113 @@ def _launch_workfile_app():
...
+        if instance.creator_identifier == "workfile":
+            continue
+
+        if (
+            instance.get("folderPath") == current_folder_path
+            and instance.get("task") == current_task
+        ):
+            continue
+
+        creator = create_context.creators[instance.creator_identifier]
```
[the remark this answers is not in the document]
BigRoy: I'm not sure how likely this case would be - because Instances would really only be listed if they were collected or created by a Creator. Could the creator identifier of the instance be customized and differ from a creator. I suppose technically it could, but it'd really just point to potential mis-use. So I'm actually fine with letting this error so we capture it and actually fix it instead of having lingering warnings for a while if it were the case.

client/ayon_nuke/api/lib.py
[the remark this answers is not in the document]
sebastianbrandhuber: I agree, this makes sense.

moonyuet (approved):
Tested successfully in my side

BigRoy (approved):

PR #269, opened 2026-06-17
fix: version imprint defaults to None for missing keys

### LINKED ISSUE #268: NukeGroupLoader: _get_imprint_data() errors for non-existing version keys

## Current Behavior
If a gizmo is loaded and the version does not have one of those keys:
https://github.com/ynput/ayon-nuke/blob/fa4adc679a797163680c102753b2f0b50d1c09c6/client/ayon_nuke/api/plugin.py#L874-L879
The load fails with a key error.

## Expected Behavior
Skip the missing key and do not error.

### Steps To Reproduce:

1. Have a published gizmo where one of the keys mentioned above is missing,
for example a custom LUT ingest where frameStart or frameEnd could not be determined.
2. Load the gizmo with the loader

### Version
ayon-nuke 0.4.10

Windows 11

### Suggested solution
Change this line
https://github.com/ynput/ayon-nuke/blob/fa4adc679a797163680c102753b2f0b50d1c09c6/client/ayon_nuke/api/plugin.py#L881
to
`data[k] = version_attributes.get(k)`

I will create a smal PR for this unless there is not a very specific reason 
why missing keys should not default to None and eventually get removed from the return dict.

### Relevant log output:

```shell

```

sebastianbrandhuber:
## Changelog Description
Missing keys in version data imprint will not result in a key error anymore and are eventually removed from the data dictionary.

More information about this [here](https://github.com/ynput/ayon-nuke/issues/268)

Fix https://github.com/ynput/ayon-nuke/issues/268

## Testing notes:
1. Have a published gizmo where one of these attributes is missing:
```
            "frameStart",
            "frameEnd",
            "handleStart",
            "handleEnd",
            "source",
            "fps"
```
for example a custom LUT ingest where frameStart or frameEnd could not be determined.
3. Load the gizmo with the loader - should not fail.


BigRoy (approved):
LGTM - `None` values are skipped below.

jakubjezek001 (approved):

jakubjezek001:
Thanks a lot!

PR #275, opened 2026-06-24
Extract Review Intermediate: do not use localized file

### LINKED ISSUE #75: Render (write) - review, uses localized files, resulting in outdated reviews on AYON

## Current Behavior
Ayon sometimes uploads an outdated review file when publishing a render.

When an artist loads in the render using the **"Read From Rendered"** button, which loads in the non published render (from the work folder) Nuke will automatically localize the files (if this behavior is set in the nuke settings), which is nice, as the playback will be way faster than over the network.

But when publishing a shot, it overwrites these files before copying it to the publish folder.

So where it goes wrong is; when rendering the review movie, it loads in the render, but Nuke uses the localized files instead of the new render. (Nuke doesn't immediately recognize that the cache has changed as the path is still the same as the old render.)

## Expected Behavior
Ayon/Nuke should not use the localized files when generating the review movie. 
So in the **read node** generated to create the review movie we should set the **"Localization Policy"** to **"off"**

### Steps To Reproduce:

_No response_

### Versions
Version of the server: 1.7 
ayon-core: 1.0.14
addon: 0.3.1

### Relevant log output:

```shell

```

moonyuet:
## Changelog Description
This PR is to make sure Extract Review Intermediate not using localized file for rendering the movie.
Resolve https://github.com/ynput/ayon-nuke/issues/75

n/a

## Testing notes:
1. Create Render with local rendering option
2. Publish
3. Adjust the frames to be rendered(Like adding grading, exposure etc.)
4. Publish again
5. Go check the reviewable in your ayon servers
6. The review movie should not be outdated version anymore.


moonyuet:
@basvanbergen would be awesome if you can take a look too.

client/ayon_nuke/api/plugin.py
```diff
@@ -1296,6 +1296,9 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
         # Read node
         r_node = nuke.createNode("Read")
         r_node["file"].setValue(self.path_in)
+        # do not use the localized files when publishing,
+        # use the original files
```
BigRoy: I wonder if we should elaborate some more?
```suggestion
        # use the original files, because Nuke may think the
        # cached localized files are still up-to-date and use them
```

BigRoy (approved):
Didn't test it myself, but code makes sense. Can @LiborBatek or @jakubjezek001  test this?

basvanbergen:
> @basvanbergen would be awesome if you can take a look too.

I would, but I’m not at a computer for the next two weeks, sorry. Let me know if it’s still needed after that.

LiborBatek (approved):
I have produced two versions while `local rendering` and each been correct and represented the changes among them...LGTM

No issues were spotted.

Image

PR #276, opened 2026-06-25
Fix avalon knob data growing on version updates by removing node reference

### LINKED ISSUE #252: avalon:node knob keeps growing bigger on every version update in Scene Inventory

## Current Behavior
Every time I change the version of a loaded container in Scene Inventory (up or down), a hidden knob called `avalon:node` gets added to the node — or if it already exists, it gets bigger. After two version changes it's about twice the size it was after one. This bloats the `.nk` file and any scene submitted to the farm.

## Expected Behavior
The `avalon:node` knob shouldn't exist at all. All the container info is already saved in the other `avalon:*` knobs (`avalon:representation`, `avalon:version`, etc.). Nothing in the codebase reads `avalon:node`, so it seems like it's being created by accident.

### Steps To Reproduce:

1. Load any product in Nuke via the AYON Loader (e.g. a render EXR sequence).
2. Open Scene Inventory and switch to an older version.
3. Switch back to the original version.
4. Run this in the Script Editor:
   ```python
   import nuke
   for n in nuke.allNodes(recurseGroups=True):
       if "avalon:node" in n.knobs():
           print(f"{n.name()}: avalon:node length = {len(n['avalon:node'].value())}")
   ```
5. Repeat steps 2–4 and run the snippet again. The number keeps doubling each time.

### Additional context:

I think the issue is in `update_container` in `api/pipeline.py`. It reads the container data using `parse_container`, but that function also adds some extra fields to the dict — including `"node"`, which is the actual Nuke node object. When that dict gets written back to the node's knobs, Nuke converts the node object to a string containing all of the node's current state. That string ends up saved as the `avalon:node` knob, and it includes a copy of itself from the previous update, so it keeps nesting.

### Version
- ayon-nuke: bug introduced in commit `978ec89f` (November 2023)

Windows
MacOS

### Relevant log output:

Knob size on a single Read node across version changes:

| State | `avalon:node` length |
|---|---|
| After initial load | NOT PRESENT |
| After first version change | 1 437 chars |
| After second version change | 2 985 chars

```json
[
  {
    "step": "load v3",
    "node": "LoadClip_exr1",
    "avalon:representation": "aaa...aaa",
    "avalon:version": 3,
    "file": "P:/studio/.../product/v003/...v003.%04d.exr",
    "avalon:node": null
  },
  {
    "step": "v3 -> v1",
    "node": "LoadClip_exr1",
    "avalon:representation": "bbb...bbb",
    "avalon:version": 1,
    "file": "P:/studio/.../product/v001/...v001.%04d.exr",
    "avalon:node": {
      "length": 1437,
      "avalon:representation": "bbb...bbb",
      "avalon:version": 1,
      "file": "P:/studio/.../product/v001/...v001.1010.exr",
      "avalon:node": null
    }
  },
  {
    "step": "v1 -> v3",
    "node": "LoadClip_exr1",
    "avalon:representation": "aaa...aaa",
    "avalon:version": 3,
    "file": "P:/studio/.../product/v003/...v003.%04d.exr",
    "avalon:node": {
      "length": 2985,
      "avalon:representation": "bbb...bbb",
      "avalon:version": 1,
      "file": "P:/studio/.../product/v003/...v003.1010.exr",
      "avalon:node": {
        "avalon:representation": "bbb...bbb",
        "avalon:version": 1,
        "file": "P:/studio/.../product/v001/...v001.1010.exr",
        "avalon:node": null
      }
    }
  }
]
```

jakubjezek001:
## Changelog Description
Fixed an issue where the avalon knob data continuously grew in size with each version update in Scene Inventory. The problem occurred because the `node` key, containing a reference to the Nuke node object, was being serialized into the avalon knob data. This caused the knob to accumulate redundant node references on every update, bloating the node's data unnecessarily.

The fix removes the `node` key from the container data before setting it into the node's avalon knob, preventing the accumulation of node references and keeping the knob data clean and minimal.

## Additional info
The change is implemented in `set_avalon_knob_data()` function in `client/ayon_nuke/api/lib.py`. The `node` key is now explicitly removed from the data dictionary before the avalon knob is populated. This is safe because the node reference is not needed to be stored in the knob - it's only used during runtime operations and should not be persisted in the node's serialized data.

## Testing notes:
1. Load a container in Nuke Scene Inventory
2. Update the container to a newer version multiple times
3. Inspect the avalon knob data on the node - it should not grow with each update
4. Verify that container functionality (updates, switches, removals) works as expected

## Dependency
Close #252

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -428,6 +428,11 @@ def set_avalon_knob_data(node, data=None, prefix="avalon:"):
     data = data or dict()
     create = OrderedDict()
 
+    # remove node key from container data before
+    # setting it into the node's avalon knob
+    if "node" in data:
+        del data["node"]
+
     tab_name = NODE_TAB_NAME
     editable = ["folderPath", "productName", "name", "namespace"]
 
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -430,8 +430,11 @@ def set_avalon_knob_data(node, data=None, prefix="avalon:"):
 
     # remove node key from container data before
     # setting it into the node's avalon knob
-    if "node" in data:
-        del data["node"]
+    data = {
+        key: value
+        for key, value in data.items()
+        if key not in ("node", "objectName")
+    }
 
     tab_name = NODE_TAB_NAME
     editable = ["folderPath", "productName", "name", "namespace"]


client/ayon_nuke/api/lib.py
```diff
@@ -428,6 +428,11 @@ def set_avalon_knob_data(node, data=None, prefix="avalon:"):
     data = data or dict()
     create = OrderedDict()
 
+    # remove node key from container data before
+    # setting it into the node's avalon knob
+    if "node" in data:
+        del data["node"]
```
iLLiCiTiT:
```suggestion
    data.pop("node", None)
```
And we probably also should pop `"objectName"`.
```suggestion
    data.pop("node", None)
    data.pop("objectName", None)
```

And maybe, because the data is shared object, we should create new dictionary
```suggestion
    data = {
        key: value
        for key, value in data.items()
        if key not in ("node", "objectName")
    }
```

BigRoy:
@moonyuet can you confirm loading and updating content in nuke still works as intended?

It should also fix this issue: https://github.com/ynput/ayon-nuke/issues/252

moonyuet (approved):
https://github.com/user-attachments/assets/1a8d3903-196a-40ae-a2fe-547a305d8e17

Tested successfully and see the video capture above

BigRoy:
@jakubjezek001 feel free to merge if ok for you like this.

BigRoy (approved):

PR #278, opened 2026-06-26
Reuse check inventory versions for loader colors

vincentullmann:
## Changelog Description
refactor the tile_color logic to be shared between loaders. Updated the `check_inventory_versions` to use the same logic as the loader for consistent color coding

before:
- each Loader re-implements the `is_latest` checks to update the node colors
- `check_inventory_versions` runs on scriptSave and uses its own logic and own color values to update the node colors
- loading an older version of a product incorrectly used the `latest` color

after:
- `plugin.NukeLoader.get_node_colors` returns a map of "container_status/categories" to colors
   - those values can be overwritten either by setting the respective `node_color_<status>` variables on the LoaderClass
   - or by overwriting `get_node_colors` and return different values
- `plugin.NukeLoader.update_node_color(self, node)` a helper method to update the tile color for a given node. Loaders should use this on `load` and `update`
- `check_inventory_versions` now uses the colors defined by each containers `get_node_colors`

Todo:
- so far I've only updated the `LoadClip` and `LoadImage` loaders as examples. If the general approach is approved I'm happy to roll out the updates to other loaders too.

## Testing notes:
1. import a product using the `LoadClip` or `LoadImage` loader
2. => confirm that the correct color (latest) was used
3. change the version to an older one
4. => confirm that the correct color was used (outdated color)
5. change it back to latest
6. => confirm that the correct color (latest) was used
7. load a outdated version of a product
8. => confirm that the correct color was used (outdated color)
9. save the workfile (this should trigger the `onScriptSave` callback: `check_inventory_versions`
10. confirm that all nodes have the correct colors


client/ayon_nuke/api/plugin.py
```diff
@@ -59,6 +65,27 @@
...
+
+def _get_library_projects() -> Generator[ProjectDict, None, None]:
+    """Get library projects."""
+    return ayon_api.get_projects(library=True, fields={"id", "name"})
+
+
+@functools.cache
+def _get_library_project_names() -> list[str]:
+    """Get library project names."""
+    return [project["name"] for project in _get_library_projects()]
```
BigRoy: Better as `set[str]` response?

client/ayon_nuke/api/plugin.py
```diff
@@ -628,6 +655,23 @@ class NukeLoader(LoaderPlugin):
...
+    node_color_latest           = color_to_int(78, 205, 37)  # 0x 4e cd 25 ff
+    node_color_outdated         = color_to_int(216, 79, 32)  # 0x d8 4f 20 ff
+
+    # library project
+    node_color_library_latest   = color_to_int(143, 132, 61) # 0x 8f 84 3d ff
+    node_color_library_outdated = color_to_int(120, 110, 50)
+
+    # other projects
+    node_color_other_project_latest = color_to_int(200, 32, 32)
+    node_color_other_project_outdated = color_to_int(180, 28, 28)
```
BigRoy: Admittedly - I think this may get to a point of having just _more colors_ where the user will not be able to distinguish what they actually mean unless very explicitly told about it.

I also don't see much reason myself why we'd want to distinguish this far - something is either latest version, or it's not. Mixing in color coding of it being from another project, or what happens to be a library project I think may just be a source of confusion.

Also, these require an additional query from the server (even though you're technically caching the names ones to optimize that issue.)

@jakubjezek001 @LiborBatek how do you feel about color coding with these variations?

client/ayon_nuke/api/plugin.py
```diff
@@ -59,6 +65,27 @@
...
 
+if typing.TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from ayon_api.typing import ProjectDict
+
+
+def _get_library_projects() -> Generator[ProjectDict, None, None]:
+    """Get library projects."""
+    return ayon_api.get_projects(library=True, fields={"id", "name"})
```
BigRoy: I'd say, let's remove this - and put the api call directly in `_get_library_project_names`

client/ayon_nuke/api/plugin.py
vincentullmann: yeah, I'm happy to combine `node_color_other_project_latest` and `node_color_other_project_outdated` into a single `node_color_other_project` and do the same for `node_color_library`

that would leave us with 6 colors rather than 8:
```py
node_color_invalid
node_color_not_found

# current project
node_color_latest
node_color_outdated

# other projects
node_color_library
node_color_other_project
```

Personally, I think it would even be fine if `node_color_other_project = node_color_not_found or node_color_invalid`, it just feels a bit off semantically.

The most important feature for us would be top be able to give library assets their own unique color, to be able to clearly distinguish them from the current project or other non-library projects.

vincentullmann: I've removed any logic regarding "library" and "other_project" colors.

So this PR now purely focuses on refactoring the node_color logic to be centralised,
as in making sure the `onSave/onLoad` callback uses the same logic as the individual loaders

BigRoy:
@vincentullmann can you resolve the linter issues?

client/ayon_nuke/api/plugin.py
```diff
@@ -628,6 +631,11 @@ class NukeLoader(LoaderPlugin):
     container_id_knob = "containerId"
     container_id = None
 
+    node_color_invalid    = color_to_int(255, 0, 0)    # 0xff0000ff
+    node_color_not_found  = color_to_int(255, 255, 0)  # 0xffff00ff
+    node_color_latest     = color_to_int(78, 205, 37)  # 0x4ecd25ff
+    node_color_outdated   = color_to_int(216, 79, 32)  # 0xd84f20ff
```
moonyuet, line 637: @BigRoy @vincentullmann Guess it might be good idea to put it into the addon settings and users can customize their own colors for these nodes?

vincentullmann: yeah, I like the idea.

On that note: When I was going trough each loader and updating the logic I noticed that, the defaults I've picked based on `LoadCLip` get overwritten often by the same colors.

here is a table showing what colors are currently being used:
| Loader | invalid | not_found | latest | outdated |
|---|---|---|---|---|
| `NukeLoader` |  |  |  |  |
| `LoadClip` | default | default | default | default |
| `LoadImage` | default | default | default | default |
| `LinkAsGroup` | default | default |  |  |
| `AlembicModelLoader` | default | default |  |  |
| `GeoImportLoader` | default | default |  |  |
| `GeoReferenceLoader` | default | default |  |  |
| `LoadOcioLookNodes` | default | default |  |  |
| `AlembicCameraLoader` | default | default |  |  |
| `FbxCameraLoader` | default | default |  |  |
| `UsdCameraLoader` | default | default |  |  |

This raises the questions:
- a) Should each loader have their own server settings?
- b) or should we unify the colors? eg.:  latest always  and outdated always  ?
- c) or group them up and give each group a set of server settings?
    for example 
    -  2D Nodes (`Clip`, `Image`)
    -  `Model` and `Geo`
    -  `*CameraLoader`s
    - not sure where `LinkAsGroup` and `LoadOcioLookNodes` would fit in

vincentullmann: Note on b):
If we unify the colors, we could simplify the code as well. Currently there is some work done to allow each loader to define its own colors, which would be unnecessary, if we'd pick that option

BigRoy: If there's need for `c)` then I think we should go for `a)` and allow customizing all separately.
If no need to differentiate between each, I think `b)` is the best option.

My vote would go for `b)` but one may like to change to 3d nodes to a different color from the read nodes (especially in the case of not-outdated) which would require `a)`. I don't have Nuke experience so hard to tell what a user would prefer there. @jakubjezek001 @rdelillo preferences?

jakubjezek001:
closing for #364

PR #279, opened 2026-06-26
Named undo blocks

vincentullmann:
## Changelog Description
wrap `load`, `update` and `delete` into named undo steps

This PR adds a new context manager which can be used to wrap functions into a single named Undo-Step.
This improves readability when users look at their undo/redo steps

### Note:
For now I've only updated the `LoadClip` loader. Once the general Idea/Implementation is approved I'm happy to roll it out to other loaders too

load
image

update
image

delete
image

## Testing notes:
1. load, update and remove a product (using the `LoadClip` loader)
2. check the undo history


client/ayon_nuke/plugins/load/load_clip.py
vincentullmann, line 1: best to view this without whitespace.
Most of the lines only have an change in indentation

client/ayon_nuke/api/command.py
```diff
@@ -19,3 +19,36 @@ def viewer_update_and_undo_stop():
         yield
     finally:
         nuke.Undo.enable()
+
+
+@contextlib.contextmanager
+def undo_step(name: str = ""):
```
BigRoy: Where does the `step` terminology come from?

If just something you came up with - can we make it `chunk` instead? Only reason is then at least it matches with `ayon-maya` and `ayon-fusion`, `ayon-silhouette`, `ayon-cinema4d`

client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -180,59 +180,59 @@ def load(self, context, name, namespace, options):
             inpanel=False
         )
```
BigRoy, line 182: Previously this also stopped the active viewer. Is it safe without that? @jakubjezek001

client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -293,11 +294,20 @@ def update(self, container, context):
...
 
         read_node = container["node"]
 
+        # update undo name
+        old_container = parse_container(read_node) or {}
+        if old_version := old_container.get("version"):
+            name = container.get("name") or read_node.name()
+            name = f"Update: {name} v{old_version} -> v{version_name}"
+            nuke.Undo.name(name)
+
```
BigRoy:
```suggestion
```

client/ayon_nuke/plugins/load/load_clip.py
```diff
@@ -425,15 +430,23 @@ def set_colorspace_to_node(
...
-            members = self.get_members(read_node)
-            nuke.delete(read_node)
-            for member in members:
-                nuke.delete(member)
+        # update undo name
+        name = container.get("name") or read_node.name()
+        if version := container.get("version"):
+            undo_name = f"Remove: {name} v{version}"
+        else:
+            undo_name = f"Remove: {name}"
```
BigRoy: Let's simplify this - I really would like to remove `version` as container data in Nuke in time anyway - it's not standard practice to have that imprinted. So let's avoid using it here.

Let's name it purely by the `name` here.

BigRoy (commented):
Nice, I like it.

Is there a reason why you only updated `load_clip.py` and not the other loaders?

client/ayon_nuke/api/command.py
vincentullmann: Ah, I wasn't aware what terminology was used in other ayon addons. renamed it to `undo_chunk` to match

client/ayon_nuke/plugins/load/load_clip.py
vincentullmann: yes, this was a side effect I was initially planning to fix

I did notice though that now its possible to upate versions whilst the player is playing back... which is a very niece use case but fun :D

https://github.com/user-attachments/assets/429a3b52-c08f-486b-9ab1-cf0d9a49ff99

BigRoy: @jakubjezek001 was there a good reason to disallow it?

vincentullmann:
> Is there a reason why you only updated `load_clip.py` and not the other loaders?

mostly to first see if the general implementation is approved, to avoid having to roll out updates (such as the `undo_step` -> `undo_chunk` rename) to multiple different files

client/ayon_nuke/plugins/load/load_clip.py
jakubjezek001: mainly to avoid cache conflicts, but I am only guessing since it is while ago when I was writing it. This is really 2019 code..

jakubjezek001 (approved):
this works well and looks good to me too

jakubjezek001:
Do you want to create new PR with implementation of plugins targeting this PR please?

BigRoy:
> Do you want to create new PR with implementation of plugins targeting this PR please?

I'd say, @vincentullmann please include it in this PR. No reason to separate it.

vincentullmann:
did a first pass on all the other plugins, replacing the `viewer_update_and_undo_stop` blocks with `@undo_chunk` decorators (and adding new ones, for any loaders which didn't had any before)

Not done yet:
- update undo-block names on Loaders based in `NukeGroupLoader`:
   for those I did add the decorator on the baseClass, so all child loaders currently would use the same generic `Load Group`, `Update Group` and `Remove Group` labels. Works but not pretty
 - update undo_chunk names based on the product name / version.
   in the `LoadClip` Example I was refining the undo chunk name from `Load Clip`  to `Load Clip: {productName} v{version}`. I'd like to avoid copy pasting the same code block into every loader, so I might add a helper method to address that (this could be a seperate pr)

BigRoy:
> * update undo-block names on Loaders based in `NukeGroupLoader`:
>   for those I did add the decorator on the baseClass, so all child loaders currently would use the same generic `Load Group`, `Update Group` and `Remove Group` labels. Works but not pretty

Actually not that problematic for me, it's cosmetics at this point.

> I'd like to avoid copy pasting the same code block into every loader, so I might add a helper method to address that (this could be a seperate pr)

Fine with me either now or separate. @jakubjezek001 thoughts?

BigRoy:
@jakubjezek001 can you review this next week please?

jakubjezek001:
@vincentullmann this is mergable, please update the branch before I can merge it. Thank you.

jakubjezek001 (approved):

PR #283, opened 2026-07-02
Make all validators optional

vincentullmann:
## Changelog Description
made `ValidateExposedKnobs`, `ValidateProxyMode`, `ValidateRenderedFrames` and `ValidateNukeWriteNode` optional

initially I was planning to only update `ValidateExposedKnobs` but I figured while at it I'd update these too.
With the PR all `ayon-nuke` validators should be optional, allowing studios to disable them if they want to replace or simply not use them in their workflows.

## Testing notes:
1. disable any of these validators (either in server settings or publish settings)
2. publish
3. confirm they didn't run

### DIFF
--- a/client/ayon_nuke/plugins/publish/validate_proxy_mode.py
+++ b/client/ayon_nuke/plugins/publish/validate_proxy_mode.py
@@ -27,7 +27,7 @@ class ValidateProxyMode(
     """Validate active proxy mode"""
 
     order = pyblish.api.ValidatorOrder
-    optional = True
+    optional = False
     label = "Validate Proxy Mode"
     hosts = ["nuke"]
     actions = [FixProxyMode]
--- a/client/ayon_nuke/plugins/publish/validate_proxy_mode.py
+++ b/client/ayon_nuke/plugins/publish/validate_proxy_mode.py
@@ -1,6 +1,9 @@
 import pyblish
 import nuke
-from ayon_core.pipeline import PublishXmlValidationError
+from ayon_core.pipeline import (
+    PublishXmlValidationError,
+    OptionalPyblishPluginMixin
+)
 
 
 class FixProxyMode(pyblish.api.Action):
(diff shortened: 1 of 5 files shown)


BigRoy:
I have no issues with exposing more validators to settings, but I feel like the default "optional" state should match what it was before in default settings.

moonyuet (approved):
The validators showed up as expected. 

Image

And can be turned on/off and they won't be triggered when being turned off as expected.

vincentullmann:
> I have no issues with exposing more validators to settings, but I feel like the default "optional" state should match what it was before in default settings.

yes, makes sense. done

client/ayon_nuke/plugins/publish/validate_rendered_frames.py
```diff
@@ -44,10 +47,14 @@ def process(self, context, plugin):
...
 
-class ValidateRenderedFrames(pyblish.api.InstancePlugin):
+class ValidateRenderedFrames(
+    OptionalPyblishPluginMixin,
+    pyblish.api.InstancePlugin
+):
     """Validates file output."""
 
     order = pyblish.api.ValidatorOrder + 0.1
+    optional = True
```
moonyuet:
```suggestion
    optional = False
```

client/ayon_nuke/plugins/publish/validate_proxy_mode.py
```diff
@@ -17,17 +20,23 @@ def process(self, context, plugin):
...
 
-class ValidateProxyMode(pyblish.api.ContextPlugin):
+class ValidateProxyMode(
+    OptionalPyblishPluginMixin,
+    pyblish.api.ContextPlugin,
+):
     """Validate active proxy mode"""
 
     order = pyblish.api.ValidatorOrder
+    optional = True
```
moonyuet:
```suggestion
    optional = False
```

PR #330, opened 2026-07-10
fix: remove camera copy/paste workaround

vincentullmann:
## Changelog Description
simplify camera load logic, by removing the copy/paste workaround

tested with an alembic camera in Nuke 15
this is an alternative to https://github.com/ynput/ayon-nuke/pull/329

## Testing notes:
1. load a camera
2. make sure the values have been applied
3. update the camera
4. make sure the new values have been applied

### DIFF
--- a/client/ayon_nuke/plugins/load/load_camera.py
+++ b/client/ayon_nuke/plugins/load/load_camera.py
@@ -75,22 +75,8 @@ class AlembicCameraLoader(load.LoaderPlugin):
                     inpanel=False,
                 )
 
-            # get the actual name of the camera node
-            # might be different if a the desired name is already in use
-            object_name = camera_node.name()
-
-            camera_node.forceValidate()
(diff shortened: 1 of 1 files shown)


moonyuet:
I tested it in my side, and I prefer this fix than [another camera-selection PR](https://github.com/ynput/ayon-nuke/pull/329). 
The only thing which might be quite annoying is the pop-up dialog to ask you whether you wanna read camera from source file and destroy it. Other than that, I think this one directly set up the versioned filepath in the camera node is a better fix
image

@jakubjezek001 @rdelillo @BigRoy thoughts?

vincentullmann:
> The only thing which might be quite annoying is the pop-up dialog to ask you whether you wanna read camera from source file and destroy it.

Ah yeah,

that popup can be supressed by enabling this checkbox
I'm currently doing that in a `LoaderHookPlugin` in our inhouse plugin
image

PR #335, opened 2026-07-21
[YN-0943]: Add slate generation as a Nuke instance attribute

### LINKED ISSUE #333: YN-0943: slate generator processing trigger

client requires to be able to trigger slate processing from nuke write node without need for nuke node in the scene 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0943]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=42d0f38afd1347caa40c374f1d154d4f)

jakubjezek001:
## Changelog Description

Adds a new **slate_gen** instance attribute ("Slate Generation", default `True`) to the Nuke writer plugin. When enabled on an instance, slate generation is triggered as part of the write workflow. The definition joins existing options (Reviewable, Farm rendering, Use range limit, Render On Farm) in one place and is consumed by both client and server sides.

## Additional info

N/A

## Testing notes:

1. Create a Nuke Write instance with Slate Generation enabled (default). Verify the attribute appears correctly in the server UI and on the client side.
2. Create another instance with Slate Generation disabled and confirm workflow behavior matches expectations.

## Dependency
Need to be tested with: https://github.com/ynput/ayon-core/pull/1965
Close https://github.com/ynput/ayon-nuke/issues/333

## Related support tickets

[YN-0943](https://support.ayon.app/projects/Active_Clients/overview/projects/Active_Clients/overview/projects/Active_Clients?project=Active_Clients&type=task&id=f2d9f886affb455a9dbaa9ecc0221cca)

### DIFF
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -459,6 +459,14 @@ class NukeWriteCreator(NukeCreator):
                     label="Review"
                 )
             )
+        if "slate_gen" in self.instance_attributes:
+            attr_defs.append(
+                BoolDef(
+                    "slate_gen",
+                    default=True,
+                    label="Slate Generation"
+                )
+            )
 
         return attr_defs
 
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -1,10 +1,10 @@
 import os
-import nuke
+
 import pyblish.api
-
 from ayon_core.pipeline import publish
-
 from ayon_nuke import api as napi
+
+import nuke  # noqa
 
 
 class CollectNukeWrites(pyblish.api.InstancePlugin,
@@ -23,12 +23,18 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
     _frame_ranges = {}
 
     def process(self, instance):
-
         # compatibility. This is mainly focused on `renders`folders which
         # were previously not cleaned up (and could be used in read notes)
         # this logic should be removed and replaced with custom staging dir
         if instance.data.get("stagingDir_persistent") is None:
             instance.data["stagingDir_persistent"] = True
+
+        # add slate family to instance.data["families"]
+        if instance.data.get("slate_gen"):
+            if instance.data.get("families") is None:
+                instance.data["families"] = []
+            instance.data["families"] = list(
+                set(instance.data["families"] + ["slate"]))
 
         group_node = instance.data["transientData"]["node"]
 
@@ -314,6 +320,9 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             "tags": []
         }
 
+        if instance.data.get("slate_gen"):
+            representation["tags"].append("slate-frame")
+
         if len(collected_frames) == 1:
             representation['files'] = collected_frames.pop()
         else:
@@ -353,6 +362,8 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             list: collected frames
         """
         if "slate" not in instance.data["families"]:
+            return collected_frames
+        if instance.data.get("slate_gen"):
             return collected_frames
 
         write_node = self._write_node_helper(instance)
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -375,6 +375,8 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         """
         if "slate" not in instance.data["families"]:
             return collected_frames
+        # slate_gen mode does not need to prepend slate frame
+        # since it will be created later on downstream
         if instance.data.get("slate_gen"):
             return collected_frames
 
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -107,10 +107,21 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         write_file_path = nuke.filename(write_node)
         output_dir = os.path.dirname(write_file_path)
 
-        instance.data["expectedFiles"] = [
+        expected_files = [
             os.path.join(output_dir, source_file)
             for source_file in collected_frames
         ]
+
+        # When slate generation is enabled, include the expected slate frame
+        # path so the farm integration knows to expect it.
+        if instance.data.get("slate_gen"):
+            first_frame, _ = self._get_frame_range_data(instance)
+            slate_first_frame = first_frame - 1
+            expected_slate_path = write_node["file"].evaluate(slate_first_frame)
+            if expected_slate_path not in expected_files:
+                expected_files.insert(0, expected_slate_path)
+
+        instance.data["expectedFiles"] = expected_files
 
     def _get_frame_range_data(self, instance):
         """Get frame range data from instance.
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -117,7 +117,8 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         if instance.data.get("slate_gen"):
             first_frame, _ = self._get_frame_range_data(instance)
             slate_first_frame = first_frame - 1
-            expected_slate_path = write_node["file"].evaluate(slate_first_frame)
+            expected_slate_path = write_node["file"].evaluate(
+                slate_first_frame)
             if expected_slate_path not in expected_files:
                 expected_files.insert(0, expected_slate_path)
 
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -30,11 +30,11 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             instance.data["stagingDir_persistent"] = True
 
         # add slate family to instance.data["families"]
-        if instance.data.get("slate_gen"):
-            if instance.data.get("families") is None:
-                instance.data["families"] = []
-            instance.data["families"] = list(
-                set(instance.data["families"] + ["slate"]))
+        if (
+            instance.data.get("slate_gen")
+            and "slate" not in instance.data["families"]
+        ):
+            instance.data["families"].append("slate")
 
         group_node = instance.data["transientData"]["node"]
 
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -373,6 +373,9 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         if (
             os.path.exists(expected_slate_path)
             or (
+                # When submitting to farm using existing frames on disk
+                # and slate generation is enabled then ensure to add
+                # the slate frame even though it may not exist yet
                 instance.data["render_target"] == "frames_farm"
                 and instance.data.get("slate_gen")
             )
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -406,6 +406,13 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         })
         self.log.info("Farm rendering ON ...")
 
+        if instance.data.get("slate_gen"):
+            write_node = self._write_node_helper(instance)
+            slate_representation_ext = instance.data.setdefault(
+                "slateRepresentationExt", [])
+            slate_representation_ext.append(
+                write_node["file_type"].value())
+
     def _get_collected_frames(self, instance):
         """Get collected frames.
 
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -112,16 +112,6 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             for source_file in collected_frames
         ]
 
-        # When slate generation is enabled, include the expected slate frame
-        # path so the farm integration knows to expect it.
-        if instance.data.get("slate_gen"):
-            first_frame, _ = self._get_frame_range_data(instance)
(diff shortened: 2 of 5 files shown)


iLLiCiTiT:
It looks like the attribute `"slate_gen"` is shown and stored to instance, but where it is used?

jakubjezek001:
> It looks like the attribute `"slate_gen"` is shown and stored to instance, but where it is used?

Yes, the code changes were not pushed.

jakubjezek001:
@copilot Check following TODO and implement it into the code.

TODO:
- [x] different frame is used if `existing frames` in write render attributes is used
- [x] farm processing is not handled and also need to be implemented

server/settings/create_plugins.py
```diff
@@ -21,6 +20,8 @@
     - Render On Farm: Adds a button on the Nuke node that will submit the
         render of the write node to the farm **without** triggering the
         regular publish logic. This is useful for quick test renders.
+    - Slate Generation: When enabled, a slate frame is generated and
+        prepended to the rendered sequence before publishing.
```
jakubjezek001, line 24:
```suggestion
    - Slate Generation: When enabled, a slate frame is generated and
        prepended to the rendered sequence before publishing. 
        Slater addon is required
```

client/ayon_nuke/api/plugin.py
```diff
@@ -459,6 +459,14 @@ def get_instance_attr_defs(self):
...
                 )
             )
+        if "slate_gen" in self.instance_attributes:
+            attr_defs.append(
+                BoolDef(
+                    "slate_gen",
+                    default=True,
+                    label="Slate Generation"
+                )
+            )
```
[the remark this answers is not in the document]
jakubjezek001: @copilot this seems to be irelevant since `collect_writes.py` is referencing it via `instance.data.get("slate_gen")`. validate that is correct

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -259,13 +267,41 @@ def _render_slate_to_sequence(self, instance):
                 write_node["use_limit"].setValue(0)
                 self.log.debug("__Setting render node limit")
 
+        # For 'existing frames' mode, the main sequence is already on disk.
+        # Connect the slate node's input to a Read node that reads the
+        # existing files so the slate is composited against already-rendered
+        # content rather than the live upstream graph.
+        temporary_nodes = []
```
BigRoy, line 275: These temporary nodes should really be handled with a context manager or alike to ensure the deletion, even if the logic in-between errors.

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -354,6 +375,8 @@ def _add_slate_frame_to_collected_frames(
         """
         if "slate" not in instance.data["families"]:
             return collected_frames
+        if instance.data.get("slate_gen"):
+            return collected_frames
```
BigRoy: This may need a comment explaining why, because `if slate-gen: then do not prepend the slate-frame` looks a bit odd without context.

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -101,11 +107,23 @@ def _set_expected_files(self, instance, collected_frames):
...
 
+        # When slate generation is enabled, include the expected slate frame
+        # path so the farm integration knows to expect it.
+        if instance.data.get("slate_gen"):
+            first_frame, _ = self._get_frame_range_data(instance)
+            slate_first_frame = first_frame - 1
+            expected_slate_path = write_node["file"].evaluate(
+                slate_first_frame)
+            if expected_slate_path not in expected_files:
+                expected_files.insert(0, expected_slate_path)
```
BigRoy: Isn't this exactly what [this](https://github.com/ynput/ayon-nuke/pull/335/changes#diff-27fe00275e66e51a967e0a35beb54de61049fd37ffbea8e1dea6917c10c4b276R381-R389) code that you're skipping was doing originally - why do we need to duplicate that logic?

client/ayon_nuke/plugins/publish/extract_slate_frame.py
jakubjezek001: @copilot implement the above comment

jakubjezek001: @BigRoy would you suggest to push it farther via context manager or this way it would be ok?

client/ayon_nuke/plugins/publish/collect_writes.py
jakubjezek001: this is not duplicity, it is executed in different scope by different case

jakubjezek001: [this ](https://github.com/ynput/ayon-nuke/pull/335/changes#diff-27fe00275e66e51a967e0a35beb54de61049fd37ffbea8e1dea6917c10c4b276R381-R389) is executed during local rendering and this code only during `existing frames via farm`

BigRoy: I understand that this code is triggered then - but, the logic that defines the slate frame seems to be identical in both. Or I am blind as to what differs.

[This code says: "This will happen later."](https://github.com/ynput/ayon-nuke/pull/335/changes#diff-27fe00275e66e51a967e0a35beb54de61049fd37ffbea8e1dea6917c10c4b276R378-R381)
For this exact same plug-in [to still add it here](https://github.com/ynput/ayon-nuke/pull/335/changes#diff-27fe00275e66e51a967e0a35beb54de61049fd37ffbea8e1dea6917c10c4b276R115-R123). That's not much later, that's the same plug-in. That looks like duplicated logic within the single plug-in that should be shared?

This really smells like something that could be simplified because they share functionality?

client/ayon_nuke/plugins/publish/extract_slate_frame.py
BigRoy: try...finally is fine.

client/ayon_nuke/plugins/publish/collect_writes.py
jakubjezek001: Yes I can see you point now more clearly. It would be good to remove the duplicity for sure.

jakubjezek001: @BigRoy would this work for you? 6541c73db5e1924ebb64d177b46d3e0e076439fa

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -359,7 +370,13 @@ def _add_slate_frame_to_collected_frames(
         expected_slate_frame = first_frame - 1
         expected_slate_path = write_node["file"].evaluate(expected_slate_frame)
 
-        if os.path.exists(expected_slate_path):
+        if (
+            os.path.exists(expected_slate_path)
+            or (
+                instance.data["render_target"] == "frames_farm"
+                and instance.data.get("slate_gen")
+            )
```
BigRoy, line 381: Can we add a comment here as to what it's solving?

Like:
```
# When submitting to farm using existing frames on disk
# and slate generation is enabled then ensure to add
# the slate frame even though it may not exist yet
```

Because just the if statement here does read a bit confusingly otherwise.

client/ayon_nuke/plugins/publish/collect_writes.py
BigRoy: I think it's better yes, if we can somehow clarify the case here: https://github.com/ynput/ayon-nuke/pull/335/changes#r3759465927

BigRoy (approved):
Didn't test - but code looks ok to me.

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -23,13 +23,19 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
...
         # this logic should be removed and replaced with custom staging dir
         if instance.data.get("stagingDir_persistent") is None:
             instance.data["stagingDir_persistent"] = True
 
+        # add slate family to instance.data["families"]
+        if instance.data.get("slate_gen"):
+            if instance.data.get("families") is None:
+                instance.data["families"] = []
+            instance.data["families"] = list(
+                set(instance.data["families"] + ["slate"]))
```
iLLiCiTiT: I guess the instances are collected by the core's default plugins, so the `"families"` will be there.
```suggestion
        if (
            instance.data.get("slate_gen")
            and "slate" not in instance.data["families"]
        ):
            instance.data["families"].append("slate")
```

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -42,13 +42,20 @@ class ExtractSlateFrame(publish.Extractor):
...
 
         if "representations" not in instance.data:
             instance.data["representations"] = []
 
         self._create_staging_dir(instance)
 
         with maintained_selection():
+            slate_node = instance.data.get("slateNode")
+            if not slate_node:
+                return
```
iLLiCiTiT:
```suggestion
        slate_node = instance.data.get("slateNode")
        if not slate_node:
            return

        if "representations" not in instance.data:
            instance.data["representations"] = []

        self._create_staging_dir(instance)

        with maintained_selection():
```

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -246,6 +253,7 @@ def _render_slate_to_sequence(self, instance):
         first_frame = instance.data["frameStartHandle"]
         last_frame = instance.data["frameEndHandle"]
         slate_first_frame = first_frame - 1
+        render_target = instance.data.get("render_target", "local")
```
iLLiCiTiT: Can actually happen that the `"render_target"` is not set?

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -259,16 +267,46 @@ def _render_slate_to_sequence(self, instance):
...
+            # render slate as sequence frame
+            nuke.execute(
+                instance.data["name"],
+                int(slate_first_frame),
+                int(slate_first_frame)
+            )
+        finally:
+            # Restore original slate node connections and remove temporary
+            # nodes even when rendering fails.
+            if slate_node:
```
iLLiCiTiT:
```suggestion
            if slate_node and original_slate_input:
```

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -259,16 +267,46 @@ def _render_slate_to_sequence(self, instance):
...
-        nuke.execute(
-            instance.data["name"],
-            int(slate_first_frame),
-            int(slate_first_frame)
-        )
-
-        # turn the frame range limit back on
-        if limit_on:
-            write_node["use_limit"].setValue(1)
+        # For 'existing frames' mode, the main sequence is already on disk.
```
iLLiCiTiT, line 271: What if the slate node is not directly connected to the same output as write, but has more nodes inbetween? Or we don't care?

server/settings/create_plugins.py
```diff
@@ -41,10 +43,8 @@ def instance_attributes_enum():
         {"value": "reviewable", "label": "Reviewable"},
         {"value": "farm_rendering", "label": "Farm rendering"},
         {"value": "use_range_limit", "label": "Use range limit"},
-        {
-            "value": "render_on_farm",
-            "label": "Render On Farm"
-        }
+        {"value": "render_on_farm", "label": "Render On Farm"},
+        {"value": "slate_gen", "label": "Slate Generation"}
```
iLLiCiTiT, line 47: What this does? Render only slate? What happens to other frames?

client/ayon_nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -259,16 +267,46 @@ def _render_slate_to_sequence(self, instance):
...
+                original_slate_input = slate_node.input(0)
+                r_node = nuke.createNode("Read")
+                r_node["file"].setValue(fpath)
+                r_node["first"].setValue(first_frame)
+                r_node["origfirst"].setValue(first_frame)
+                r_node["last"].setValue(last_frame)
+                r_node["origlast"].setValue(last_frame)
+                r_node["colorspace"].setValue(instance.data["colorspace"])
+                slate_node.setInput(0, r_node)
+                temporary_nodes.append(r_node)
```
iLLiCiTiT:
```suggestion
        slate_node = instance.data.get("slateNode")
        if render_target == "frames" and slate_node:
            fpath = instance.data["path"]
            original_slate_input = slate_node.input(0)
            r_node = nuke.createNode("Read")
            r_node["file"].setValue(fpath)
            r_node["first"].setValue(first_frame)
            r_node["origfirst"].setValue(first_frame)
            r_node["last"].setValue(last_frame)
            r_node["origlast"].setValue(last_frame)
            r_node["colorspace"].setValue(instance.data["colorspace"])
            slate_node.setInput(0, r_node)
            temporary_nodes.append(r_node)
```

server/settings/create_plugins.py
jakubjezek001: this only activates Slate generation workflow, So it is adding s`slate_gen` key to instance.data with `True` value. Then CollectWrites is processing it.

client/ayon_nuke/plugins/publish/extract_slate_frame.py
jakubjezek001: it needs to be directly above write node

client/ayon_nuke/plugins/publish/extract_slate_frame.py
jakubjezek001:
```suggestion
        render_target = instance.data["render_target"]
```

jakubjezek001: good point

server/settings/create_plugins.py
iLLiCiTiT: Oh, it is a list for multiselection, nevermind.
```suggestion
        {"value": "slate_gen", "label": "Generate Slate"}
```

iLLiCiTiT (approved):
Looks ok, didn't test.

Issue #211, opened 2026-03-11
Split Nuke callbacks defined by `NukeHost` and rework nuke startup


rdelillo:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior

Current behavior make Nuke setup some callbacks and operations through `NukeHost` when Nuke start in UI mode:
- https://github.com/ynput/ayon-nuke/blob/develop/client/ayon_nuke/api/pipeline.py#L159
- https://github.com/ynput/ayon-nuke/blob/develop/client/ayon_nuke/api/lib.py#L2478

These operations affects colorspace setup, resolution, frame-range...

## Expected Behavior
We need to triage those operations between:
- those who can happen on the farm (non-GUI mode) and those who are user-specific
- those we want to happen at nuke start-up and those which need to be re-applied on each workfile opening
- those who are mandatory and those we want to make optional and configured via settings
- those who happens silently vs those who require user confirmation

Once this is done, we'd need to rework `NukeHost` for this.

Some work was already discussed :
- Split UI and none-UI callback entry point: https://github.com/ynput/ayon-nuke/pull/206
- Disable callbacks via plugin settings: https://github.com/ynput/ayon-nuke/pull/193

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

moonyuet:
I take a look the PRs as stated above, these PRs already implemented to split nuke callbacks for the settings (like placeholders system in workfile builder, set context settings etc). And if I am not wrong, the NukeHost is already reworked to ensure Nuke can be installed in terminal mode. (Maybe there is some missing bits or?)
My only concern is opening the autosave files (It will have the popup).
So I have made a PR to make sure the popup won't be shown at the headless model and used @rdelillo script to test: https://github.com/ynput/ayon-nuke/pull/332

~There is also another concern: building workfile during application launch. I will do some test and update you guys.~ it works okay.

moonyuet:
We need to decouple the callback not only context setting, but the resolution, colorspace and frame range setting explicitly.

PR #336, opened 2026-07-23
Add the options to enable context and its related settings

moonyuet:
## Changelog Description
This PR is to add the options to enable and its related settings through the addon settings.
Throughout this PR, users can also choose whether enabling to set context settings during callbacks.

Marked it as draft, we need to discuss whether better designs for the setting needed for this.

## Testing notes:
1. Launch Nuke
2. Toggle the settings from `ayon+settings://nuke/general/settings_to_apply` in AYON server

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2089,9 +2089,6 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
     def set_colorspace(self):
         """Setting colorspace following presets
         """
-        if not self._settings_to_apply.get("colorspace_settings", True):
-            log.info("Colorspace settings are disabled in project settings.")
-            return
         # get imageio
         nuke_colorspace = get_nuke_imageio_settings()
 
@@ -2123,10 +2120,6 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def reset_frame_range_handles(self):
         """Set frame range to current folder."""
-        if not self._settings_to_apply.get("frame_range_settings", True):
-            log.info("Frame range settings are disabled in project settings.")
-            return
-
         if "attrib" not in self._task_entity:
             msg = "Task {} doesn't have set any 'attrib'".format(
                 self._context_label
@@ -2197,9 +2190,6 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def reset_resolution(self):
         """Set resolution to project resolution."""
-        if not self._settings_to_apply.get("resolution_settings", True):
-            log.info("Resolution settings are disabled in project settings.")
-            return
         log.info("Resetting resolution")
         project_name = get_current_project_name()
         task_attributes = self._task_entity["attrib"]
@@ -2272,11 +2262,8 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
                 "{name}".format(**kwargs)
             )
 
-    def set_context_settings(self, enabled_on_callback=True):
+    def set_context_settings(self):
         """Apply context-related settings for script creation/load events."""
-        if not enabled_on_callback:
-            log.info("Context settings are disabled in project settings.")
-            return
         self.reset_resolution()
         self.reset_frame_range_handles()
         # add colorspace menu item
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2120,6 +2120,7 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def reset_frame_range_handles(self):
         """Set frame range to current folder."""
+
         if "attrib" not in self._task_entity:
             msg = "Task {} doesn't have set any 'attrib'".format(
                 self._context_label
@@ -2263,7 +2264,6 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
             )
 
     def set_context_settings(self):
-        """Apply context-related settings for script creation/load events."""
         self.reset_resolution()
         self.reset_frame_range_handles()
         # add colorspace menu item
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -1562,6 +1562,10 @@ class WorkfileSettings(object):
             project_name, self._folder_path, self._task_name, "nuke"
         )
         self.formatting_data = context_data
+        self._settings_to_apply = (
+            get_project_settings(project_name)["nuke"]["general"].get(
+                "settings_to_apply", {})
+        )
 
     def get_nodes(self, nodes=None, nodes_filter=None):
 
@@ -2089,6 +2093,9 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
     def set_colorspace(self):
         """Setting colorspace following presets
         """
+        if not self._settings_to_apply.get("colorspace_settings", True):
+            log.info("Colorspace settings are disabled in project settings.")
+            return
         # get imageio
         nuke_colorspace = get_nuke_imageio_settings()
 
@@ -2120,6 +2127,9 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def reset_frame_range_handles(self):
         """Set frame range to current folder."""
+        if not self._settings_to_apply.get("frame_range_settings", True):
+            log.info("Frame range settings are disabled in project settings.")
+            return
 
         if "attrib" not in self._task_entity:
             msg = "Task {} doesn't have set any 'attrib'".format(
@@ -2191,6 +2201,9 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def reset_resolution(self):
         """Set resolution to project resolution."""
+        if not self._settings_to_apply.get("resolution_settings", True):
+            log.info("Resolution settings are disabled in project settings.")
+            return
         log.info("Resetting resolution")
         project_name = get_current_project_name()
         task_attributes = self._task_entity["attrib"]
@@ -2263,7 +2276,14 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
                 "{name}".format(**kwargs)
             )
 
-    def set_context_settings(self):
+    def set_context_settings(self, enabled_on_callback=True):
+        """Apply context-related settings for script creation/load events."""
+        if not (
+            self._settings_to_apply.get("context_settings", True)
+            or enabled_on_callback
+        ):
+            log.info("Context settings are disabled in project settings.")
+            return
         self.reset_resolution()
         self.reset_frame_range_handles()
         # add colorspace menu item
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -1562,10 +1562,6 @@ class WorkfileSettings(object):
             project_name, self._folder_path, self._task_name, "nuke"
         )
         self.formatting_data = context_data
-        self._settings_to_apply = (
-            get_project_settings(project_name)["nuke"]["general"].get(
-                "settings_to_apply", {})
-        )
 
     def get_nodes(self, nodes=None, nodes_filter=None):
 
@@ -2278,10 +2274,7 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
 
     def set_context_settings(self, enabled_on_callback=True):
         """Apply context-related settings for script creation/load events."""
-        if not (
-            self._settings_to_apply.get("context_settings", True)
-            or enabled_on_callback
-        ):
+        if not enabled_on_callback:
             log.info("Context settings are disabled in project settings.")
             return
         self.reset_resolution()
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -834,16 +834,6 @@ def get_view_process_node():
         return duplicate_node(ipn_node)
 
 
-def on_script_load():
-    """Callback for ffmpeg support"""
-    if nuke.env["LINUX"]:
-        nuke.tcl('load ffmpegReader')
-        nuke.tcl('load ffmpegWriter')
-    else:
-        nuke.tcl('load movReader')
-        nuke.tcl('load movWriter')
-
-
 def check_inventory_versions():
     """Update loaded container nodes' colors based on version state.
 
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -1523,7 +1523,12 @@ class WorkfileSettings(object):
 
     """
 
-    def __init__(self, root_node=None, nodes=None, **kwargs):
+    def __init__(
+            self,
+            root_node=None,
+            nodes=None,
+            project_settings=None,
+            **kwargs):
         project_entity = kwargs.get("project")
         if project_entity is None:
             project_name = get_current_project_name()
@@ -1552,6 +1557,18 @@ class WorkfileSettings(object):
             project_name, self._folder_path, self._task_name, "nuke"
         )
         self.formatting_data = context_data
+        self._project_setting = project_settings
+
+    @property
+    def project_settings(self) -> dict:
+        """Get project settings
+
+        Returns:
+            dict: project settings for the current project
+        """
+        if not self._project_setting:
+            self._project_setting = get_project_settings(self._project_name)
+        return self._project_setting
 
     def get_nodes(self, nodes=None, nodes_filter=None):
 
@@ -1645,7 +1662,9 @@ class WorkfileSettings(object):
             imageio_host (dict): host colorspace configurations
 
         """
-        config_data = get_current_context_imageio_config_preset()
+        config_data = get_current_context_imageio_config_preset(
+            project_settings=self.project_settings
+        )
 
         workfile_settings = imageio_host["workfile"]
         color_management = workfile_settings["color_management"]
@@ -1972,9 +1991,7 @@ Reopening Nuke should synchronize these paths and resolve any discrepancies.
             # This ensures that any values overwritten by the user is
             # not changed by the colorspace knobs set.
             colorspace_knobs = nuke_imageio_writes["knobs"]
-            all_create_settings = get_project_settings(
-                Context.project_name,
-            )["nuke"]["create"]
+            all_create_settings = self.project_settings["nuke"]["create"]
             plugin_names_mapping = {
                 "create_write_image": "CreateWriteImage",
                 "create_write_prerender": "CreateWritePrerender",
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -226,11 +226,17 @@ def add_nuke_callbacks(project_settings: dict = None):
         project_settings = get_current_project_settings()
 
     nuke_settings = project_settings["nuke"]
+    settings_to_apply = nuke_settings["general"].get("settings_to_apply", {})
+
     workfile_settings = WorkfileSettings()
 
     # Set context settings.
     nuke.addOnCreate(
-        workfile_settings.set_context_settings, nodeClass="Root")
+        lambda: workfile_settings.set_context_settings(
+            enabled_on_callback=settings_to_apply.get(
+                "context_settings_on_script_create", True
+        )), nodeClass="Root"
+    )
 
     # adding favorites to file browser
     nuke.addOnCreate(workfile_settings.set_favorites, nodeClass="Root")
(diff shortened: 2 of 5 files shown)


moonyuet:
@rdelillo @BigRoy I split the callbacks regarding to https://github.com/ynput/ayon-nuke/issues/211. I know there is another PR https://github.com/ynput/ayon-nuke/pull/193 which does the same thing on the context setting, and I will merge the changes (mainly for the context setting) once it is merged.

client/ayon_nuke/api/lib.py
```diff
@@ -2089,6 +2089,9 @@ def set_reads_colorspace(self, read_clrs_inputs: list[dict[str, str]]):
     def set_colorspace(self):
         """Setting colorspace following presets
         """
+        if not self._settings_to_apply.get("colorspace_settings", True):
```
BigRoy: `_settings_to_apply` doesn't exist on this class? 🤔

client/ayon_nuke/api/pipeline.py
```diff
@@ -226,11 +226,17 @@ def add_nuke_callbacks(project_settings: dict = None):
...
     workfile_settings = WorkfileSettings()
 
     # Set context settings.
     nuke.addOnCreate(
-        workfile_settings.set_context_settings, nodeClass="Root")
+        lambda: workfile_settings.set_context_settings(
+            enabled_on_callback=settings_to_apply.get(
+                "context_settings_on_script_create", True
+        )), nodeClass="Root"
+    )
```
BigRoy: Can't we just **not** register it if irrelevant?

```
if settings_to_apply.get("context_settings_on_script_create", True):
    nuke.addOnCreate(...)
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -246,7 +252,13 @@ def add_nuke_callbacks(project_settings: dict = None):
...
 
     # set apply all workfile settings on script load and save
-    nuke.addOnScriptLoad(WorkfileSettings().set_context_settings)
+    nuke.addOnScriptLoad(
+        lambda: workfile_settings.set_context_settings(
+            enabled_on_callback=settings_to_apply.get(
+                "context_settings_on_script_open", True
+            )
+        )
+    )
```
BigRoy: Can't we just **not** register it if irrelevant?

```
if settings_to_apply.get("context_settings_on_script_open", True):
    nuke.addOnScriptLoad(...)
```

server/settings/general.py
```diff
@@ -26,14 +50,24 @@ class MenuShortcut(BaseSettingsModel):
 
 class GeneralSettings(BaseSettingsModel):
     """Nuke general project settings."""
-
+    settings_to_apply: SettingsToApply = SettingsField(
```
BigRoy: The key name is way to general :')

Settings to apply, to what? when? etc. 

It'd almost need to just be `workfile`? and I'm also thinking whether we should take it out of `general`?

- `workfile`
- description: Apply workfile settings based on current context)
- title: Workfile Settings

@jakubjezek001 @rdelillo thoughts?

client/ayon_nuke/api/lib.py
```diff
@@ -2191,6 +2197,9 @@ def reset_frame_range_handles(self):
 
     def reset_resolution(self):
         """Set resolution to project resolution."""
+        if not self._settings_to_apply.get("resolution_settings", True):
+            log.info("Resolution settings are disabled in project settings.")
+            return
```
BigRoy: I'd argue that the if statement should be in the `set_context_settings` method. That is what set the 'settings based on current context'

But calling `reset_resolution` should really reset things, regardless of the settings.

client/ayon_nuke/api/lib.py
```diff
@@ -2263,7 +2272,11 @@ def make_format_string(self, **kwargs):
                 "{name}".format(**kwargs)
             )
 
-    def set_context_settings(self):
+    def set_context_settings(self, enabled_on_callback=True):
+        """Apply context-related settings for script creation/load events."""
+        if not enabled_on_callback:
+            log.info("Context settings are disabled in project settings.")
+            return
```
BigRoy: This makes no sense. if `enabled_on_callback=False` just not call this method... or even, don't even initialize `WorkfileSettings()` instance to begin with if it's not needed.

BigRoy (changes requested):

client/ayon_nuke/api/lib.py
moonyuet: [351e481](https://github.com/ynput/ayon-nuke/pull/336/commits/351e481116bb97b67a298d707fef5f064c2b2e09)

client/ayon_nuke/api/lib.py
moonyuet: [351e481](https://github.com/ynput/ayon-nuke/pull/336/commits/351e481116bb97b67a298d707fef5f064c2b2e09)

client/ayon_nuke/api/pipeline.py
moonyuet: [351e481](https://github.com/ynput/ayon-nuke/pull/336/commits/351e481116bb97b67a298d707fef5f064c2b2e09)

client/ayon_nuke/api/pipeline.py
moonyuet: [351e481](https://github.com/ynput/ayon-nuke/pull/336/commits/351e481116bb97b67a298d707fef5f064c2b2e09)

client/ayon_nuke/api/lib.py
moonyuet: [351e481](https://github.com/ynput/ayon-nuke/pull/336/commits/351e481116bb97b67a298d707fef5f064c2b2e09)

server/settings/general.py
rdelillo: I agree that `settings_to_apply` is too generic and that can be moved out of `General`.
Also Nuke refer to those as [`callbacks`](https://learn.foundry.com/nuke/developers/170/pythondevguide/callbacks.html) so I'd somehow suggest to have this word appear in the title somehow.

Maybe something like:
```python
class WorkfileCallbacks(BaseSettingsModel):
    """Callbacks to apply in Nuke."""
    on_script_create: bool = SettingsField(
        default=True,
        title="On Script Create",
        description="Apply callbacks from AYON context on new workfile",
    )
    on_script_open: bool = SettingsField(
        default=True,
        title="On Script Open",
        description="Apply callbacks from AYON context on loaded workfile.",
    )
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -226,11 +226,15 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    settings_to_apply = nuke_settings["general"].get("settings_to_apply", {})
+
     workfile_settings = WorkfileSettings()
 
     # Set context settings.
-    nuke.addOnCreate(
-        workfile_settings.set_context_settings, nodeClass="Root")
+    if settings_to_apply.get("context_settings_on_script_create", True):
+        nuke.addOnCreate(
+            workfile_settings.set_context_settings, nodeClass="Root"
```
rdelillo: `workfile_settings.set_context_settings` does actually [3 callbacks all at once](https://github.com/ynput/ayon-nuke/blob/088cbb887d50af81ad5ed93adb179aa5240ba631/client/ayon_nuke/api/lib.py#L2266).
Part of this PR, I'd suggest we implement them as 3 different callbacks handled individually through their own settings.

Something like:
```python
on_script_create_settings = settings_to_apply["on_script_create"]
for setting_key, func in (
    ("resolution_from_context", workfile_settings.reset_resolution),
    ("range_from_context", workfile_settings.reset_frame_range_handles),
    ("colorspace_from_context", workfile_settings.set_colorspace),
):
    if on_script_create_settings .get(setting_key, True):
        nuke.addOnCreate(func, nodeClass="Root")
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -246,7 +250,8 @@ def add_nuke_callbacks(project_settings: dict = None):
     nuke.addOnScriptSave(check_inventory_versions)
 
     # set apply all workfile settings on script load and save
-    nuke.addOnScriptLoad(WorkfileSettings().set_context_settings)
+    if settings_to_apply.get("context_settings_on_script_open", True):
+        nuke.addOnScriptLoad(workfile_settings.set_context_settings)
```
rdelillo: Same, I would favor we expose 3 settings instead of one global.

server/settings/general.py
moonyuet: [9d9037d](https://github.com/ynput/ayon-nuke/pull/336/commits/9d9037d7d108342d3c47594b67a2030d25cabfb2)

client/ayon_nuke/api/pipeline.py
moonyuet: [c8c7791](https://github.com/ynput/ayon-nuke/pull/336/commits/c8c7791864cac69a9e6048720cdc079ccc876a8e)

client/ayon_nuke/api/pipeline.py
moonyuet: [c8c7791](https://github.com/ynput/ayon-nuke/pull/336/commits/c8c7791864cac69a9e6048720cdc079ccc876a8e)

client/ayon_nuke/api/pipeline.py
```diff
@@ -226,11 +226,19 @@ def add_nuke_callbacks(project_settings: dict = None):
...
-    # Set context settings.
-    nuke.addOnCreate(
-        workfile_settings.set_context_settings, nodeClass="Root")
+    # Set all workfile settings.
+    on_script_create_settings = workfile_callback_settings["on_script_create"]
+    for setting_key, func in (
+        ("resolution_from_context", workfile_settings.reset_resolution),
+        ("range_from_context", workfile_settings.reset_frame_range_handles),
+        ("colorspace_from_context", workfile_settings.set_colorspace),
+    ):
```
BigRoy: So, discussing recently with @iLLiCiTiT there's a lurking bug here with `WorkfileSettings()` being predefined here instead of getting created when called.

I think what we should do instead, is attach a singular callback to Nuke that sets the context settings instead of three.

```py
def on_root_create():

    project_settings = get_project_settings(...)
    ...
    on_script_create_settings = workfile_callback_settings["on_script_create"]
    if any(on_script_create_settings.values()):
        workfile_settings = WorkfileSettings(project_settings=project_settings)
        if on_script_create_settings.get("resolution_from_context", True):
            workfile_settings.reset_resolution()
        ... # etc

nuke.addOnCreate(on_root_create, nodeClass="root")
```

Why?

Because currently the WorkfileSettings gets initialized on app launch, which means the project settings and context it has internally always remains that context it got initialized to, instead of being live to the context I'm moving too.

_Preferably we'd drop the whole WorkfileSettings class because it's redundant - but I can refactor that later outside of this PR._

Does that make sense?

BigRoy: Same for on script load by the way.

BigRoy (changes requested):

client/ayon_nuke/api/pipeline.py
moonyuet: I have adjusted the setting in [ed7dbc7](https://github.com/ynput/ayon-nuke/pull/336/commits/ed7dbc7235cfd2fcfc87b270f0685e4bfa0e2677) and refactor the callbacks in [1dd0554](https://github.com/ynput/ayon-nuke/pull/336/commits/1dd05545ce634f52264d3ccb9f72784206c07a3c). I integrate all the functions for the same callback into respective functions(on_root_create and on_root_load).

>Preferably we'd drop the whole WorkfileSettings class because it's redundant - but I can refactor that later outside > of this PR.

I agree - Functions such as `set_resolution` and `set_frame_range` can be much simpler and refactoring can help to increase readability. But need some readings on that class as it appears to be quite significant for setting colorspace.

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +243,54 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    # set apply all workfile settings on script load and save
+    project_settings = get_current_project_settings()
+    on_script_create_settings = (
+        project_settings["nuke"]
+                        ["workfile_callbacks"]
+                        ["on_script_create"]
+    )
+
+    if any(on_script_create_settings.values()):
+        workfile_settings = WorkfileSettings(project_settings=project_settings)
```
BigRoy: `WorkfileSettings` doesn't currently take `project_settings` as input (for whatever reason it does take `**kwargs` so it's hiding that it doesn't use it?)

Can you add support for the `project_settings=None` argument (defaulting to None)

- store `_project_settings` on init?
- add a `@property` for `project_settings` and do: `if self._project_settings is None: get_project_settings` and otherwise return the cached state? The property is just so that we only do the project settings query if we need to in any of the `set` calls.
- `set_root_colorspace` should get `get_current_context_imageio_config_preset` with `self.project_settings` 
- replace `get_project_settings` in `set_writes_colorspace` with usage of `self.project_settings`

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +243,54 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    # adding favorites to file browser
+    workfile_settings.set_favorites()
+    # template builder callbacks
+    start_workfile_template_builder()
+
+
+def on_root_load() -> None:
+    """Callback function for on script load."""
+    # fix ffmpeg settings on script
+    on_script_load()
```
BigRoy: Can we just move the logic of `on_script_load` into here and remove it from `lib.py`? it's way too specifically named, it's not really a library function.

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +243,54 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    # set checker for last versions on loaded containers
+    check_inventory_versions()
```
BigRoy: Should this actually happen in headless mode I wonder - sounds very much more of a running in GUI thing?

Let's not change behavior in this PR.

server/settings/workfile_callbacks.py
```diff
@@ -0,0 +1,48 @@
...
+    frame_range_from_context: bool = SettingsField(
+        default=True,
+        title="Frame Range from Context",
+        description="Set frame range from context on new workfile.",
+    )
+    colorspace_from_context: bool = SettingsField(
+        default=True,
+        title="Colorspace from Context",
+        description="Set colorspace from context on new workfile.",
+    )
```
BigRoy: Would it make sense to shorten these keys, to just be `resolution` `frame_range` and `colorspace`. The whole class is `ContextSettings`. Seems a bit verbose to double that up maybe?


Or name them `set_resolution`, `set_frame_range` so that it reads as:

```
on_script_create.set_resolution
```
etc.

BigRoy (commented):
Nice! Some small notes left.

client/ayon_nuke/api/pipeline.py
moonyuet: [a2c07d8](https://github.com/ynput/ayon-nuke/pull/336/commits/a2c07d87406f69608550beca6af9d38a48bf08e6)

client/ayon_nuke/api/pipeline.py
moonyuet: [a2c07d8](https://github.com/ynput/ayon-nuke/pull/336/commits/a2c07d87406f69608550beca6af9d38a48bf08e6)

server/settings/workfile_callbacks.py
moonyuet: [ff3df2e](https://github.com/ynput/ayon-nuke/pull/336/commits/ff3df2e0f0bd5e9824d3c8c6b04d918913352816)

client/ayon_nuke/api/pipeline.py
moonyuet: [875b4ee](https://github.com/ynput/ayon-nuke/pull/336/commits/875b4eef9a2217edbd2e075587c426f4ddf80122)

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +244,58 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    if any(on_script_create_settings.values()):
+        workfile_settings = WorkfileSettings(project_settings=project_settings)
+        if on_script_create_settings.get("set_resolution", True):
+            workfile_settings.reset_resolution()
+        if on_script_create_settings.get("set_frame_range", True):
+            workfile_settings.reset_frame_range_handles()
+        if on_script_create_settings.get("set_colorspace", True):
+            workfile_settings.set_colorspace()
+    # adding favorites to file browser
+    workfile_settings.set_favorites()
```
iLLiCiTiT: Solves these issues:
- Don't use `get_current_project_settings` -> marked as deprecated for 3 years.
- Variable `workfile_settings`, which is alway used for `set_favorites`, is actually always defined.
- Explicit key access to settings to avoid possible missed changes in settings models.
- Cleaner code and removed unnecessary validation by dictionary values, which might be confusing, at the end you always will get the settings, just use them.
```suggestion
    workfile_settings = WorkfileSettings()
    on_script_create_settings = (
        workfile_settings.project_settings
        ["nuke"]
        ["workfile_callbacks"]
        ["on_script_create"]
    )
    
    if on_script_create_settings["set_resolution"]:
        workfile_settings.reset_resolution()

    if on_script_create_settings["set_frame_range"]:
        workfile_settings.reset_frame_range_handles()

    if on_script_create_settings["set_colorspace"]:
        workfile_settings.set_colorspace()

    # adding favorites to file browser
    workfile_settings.set_favorites()
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +244,58 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    )
+
+    if any(on_script_load_settings.values()):
+        workfile_settings = WorkfileSettings(project_settings=project_settings)
+        if on_script_load_settings.get("set_resolution", True):
+            workfile_settings.reset_resolution()
+        if on_script_load_settings.get("set_frame_range", True):
+            workfile_settings.reset_frame_range_handles()
+        if on_script_load_settings.get("set_colorspace", True):
+            workfile_settings.set_colorspace()
```
iLLiCiTiT:
```suggestion
    workfile_settings = WorkfileSettings()
    on_script_load_settings = (
        workfile_settings.project_settings
        ["nuke"]
        ["workfile_callbacks"]
        ["on_script_load"]
    )

    if on_script_load_settings["set_resolution"]:
        workfile_settings.reset_resolution()

    if on_script_load_settings["set_frame_range"]:
        workfile_settings.reset_frame_range_handles()

    if on_script_load_settings["set_colorspace"]:
        workfile_settings.set_colorspace()
```

server/settings/workfile_callbacks.py
```diff
@@ -0,0 +1,48 @@
+from ayon_server.settings import BaseSettingsModel, SettingsField
+
```
iLLiCiTiT, line 2:
```suggestion


```

rdelillo (approved):
Approving this one cause it works as expected and it's very convenient to be able to switch callbacks individually like that 💪 
Tested on Nuke 17.0, I disabled callbacks on create and enabled on open.

Small Notes:
* Need to fix the lint
* Is it a problem to always register 1 big function as callback that read the settings vs add granular functions as callbacks based on settings ? I don't see any issue and maybe safer for now, but just want to mention it.

moonyuet:
> * Is it a problem to always register 1 big function as callback that read the settings vs add granular functions as callbacks based on settings ? I don't see any issue and maybe safer for now, but just want to mention it.

Not sure if there is any reason behind adding those functions explicitly as callbacks before. But in general(like in other hosts), we added the callback with its related function (new_callback would trigger the `on_new` function which did workfile template build, set context settings etc.). I think the nuke host should align with what we have done in terms of callback if possible.

BigRoy:
> * Is it a problem to always register 1 big function as callback that read the settings vs add granular functions as callbacks based on settings ? I don't see any issue and maybe safer for now, but just want to mention it.

A problem in which way? I don't really see any issues with it either way. I prefer myself to just register a single thing to the DCC and then keep things in AYON land where we control it our way instead. If any of these ever need to run in a certain order, etc. we then know which order it's running in - if we register them with Nuke, I wouldn't know for example.

client/ayon_nuke/api/pipeline.py
```diff
@@ -256,6 +244,61 @@ def add_nuke_callbacks(project_settings: dict = None):
...
+    if on_script_create_settings["set_colorspace"]:
+        workfile_settings.set_colorspace()
+
+    # adding favorites to file browser
+    workfile_settings.set_favorites()
+    # template builder callbacks
+    start_workfile_template_builder()
+
+
+def on_root_load() -> None:
```
BigRoy: This should be named `on_script_load`.

BigRoy (approved):
LGTM - just refactor `on_root_load` -> `on_script_load`.

client/ayon_nuke/api/pipeline.py
moonyuet: [7592e16](https://github.com/ynput/ayon-nuke/pull/336/commits/7592e160d8dd4b9ab4a1d5b9e1fa84849ffe5200)

PR #338, opened 2026-07-23
don't force evaluation of whole comp tree when searching for downstre…

### LINKED ISSUE #337: CollectWrites causes Nuke to freeze and crash

## Current Behavior
When publishing, process seems stuck at `CollectWrites`, Nuke freezes and crashes after a while

## Expected Behavior
I would expect that `CollectWrites` doesn't block the publish process and it finishes successfully.

### Steps To Reproduce:

I haven't been able to reproduce it consistently as it happens randomly but I just launch Nuke and publish ( in a more or less complex scene). 

### Version
AYON Server: `1.15.13+202607101711`.
AYON Core: `1.9.5`.
AYON Nuke: `0.4.8`. 

Linux - Rocky8

### Nuke version 

`14.1.4`

### Relevant log output:

```shell
[53865:53865:20260723,124152.223114:ERROR elf_dynamic_array_reader.h:64] tag not found
[53865:53865:20260723,124152.251915:WARNING process_reader_linux.cc:144] no stack mapping
[53865:53865:20260723,124152.263334:ERROR file_io_posix.cc:144] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq: No such file or directory (2)
[53865:53865:20260723,124152.263373:ERROR file_io_posix.cc:144] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq: No such file or directory (2)
nuke_14.1.sh: line 27: 53846 Segmentation fault      (`core' dumped) Nuke14.1v4/Nuke14.1 "$@"
```

JoseCovaOS:
This PR fixes #337 

## Changelog Description

Fixed an issue where publishing complex Nuke scenes could freeze or crash during Collect Writes. Searching for downstream TimeWarp nodes no longer forces evaluation of the node graph, including in scenes that contain no TimeWarp nodes.

Collect Writes searches downstream dependencies to determine whether TimeWarp metadata needs to be included in the published version data.

The previous implementation called `nuke.dependentNodes()` twice because the first call could return no results. These calls could force Nuke to evaluate the node graph while searching for a TimeWarp. In sufficiently complex scenes, this evaluation could cause Nuke to freeze and eventually crash—even when the composition contained no TimeWarp nodes.

The lookup now calls `Node.dependent()` from the publish instance node with `forceEvaluate=False`. This allows Collect Writes to inspect the available dependencies without triggering graph evaluation. If a TimeWarp is found, its lookup values continue to be collected as before; if none is found, publishing proceeds normally.

The obsolete `_find_downstream_time_warp_node()` helper and its repeated dependency queries were removed.

## Testing notes:

1. Launch Nuke.
2. Open a complex scene containing an AYON render, prerender, or image publish instance, but no TimeWarp nodes.
3. Publish the instance.
4. Confirm that Collect Writes completes without Nuke freezing or crashing.
5. Confirm that the publish finishes successfully.
6. Optionally repeat the test several times because the original issue occurred intermittently.
7. As a regression test, publish a scene with a downstream TimeWarp and confirm that the publish succeeds and its retime information is retained.

### DIFF
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -187,16 +187,10 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             "colorspace": colorspace
         }
 
-        time_warp_node = next(
-            (
-                dependent_node
-                for dependent_node in instance.data["transientData"][
-                    "node"
-                ].dependent(nuke.INPUTS, forceEvaluate=False)
-                if dependent_node.Class() == "TimeWarp"
-            ),
-            None,
+        time_warp_node = _find_downstream_time_warp_node(
+            instance.data["transientData"]["node"]
         )
+
         if time_warp_node:
             time_warp_dict = {
                 "Class": time_warp_node.Class(),
@@ -435,3 +429,8 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             collected_frames,
             first_frame
         )
+
+def _find_downstream_time_warp_node(start_node):
+    for node in start_node.dependent(nuke.INPUTS, forceEvaluate=False):
+        if node.Class() == "TimeWarp":
+            return node
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -187,8 +187,15 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             "colorspace": colorspace
         }
 
-        time_warp_node = _find_downstream_time_warp_node(
-            instance.data["transientData"]["node"]
+        time_warp_node = next(
+            (
+                dependent_node
+                for dependent_node in instance.data["transientData"][
+                    "node"
+                ].dependent(nuke.INPUTS, forceEvaluate=False)
+                if dependent_node.Class() == "TimeWarp"
+            ),
+            None,
         )
         if time_warp_node:
             time_warp_dict = {
@@ -428,11 +435,3 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
             collected_frames,
             first_frame
         )
-
-def _find_downstream_time_warp_node(start_node):
-    # HACK: no idea why calling `dependentNodes` the first time
-    # seems to always return nothing.
-    nuke.dependentNodes(nuke.INPUTS, [start_node])
-    for node in nuke.dependentNodes(nuke.INPUTS, [start_node]):
-        if node.Class() == "TimeWarp":
-            return node
--- a/client/ayon_nuke/plugins/publish/collect_writes.py
+++ b/client/ayon_nuke/plugins/publish/collect_writes.py
@@ -190,7 +190,6 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
         time_warp_node = _find_downstream_time_warp_node(
             instance.data["transientData"]["node"]
         )
-
         if time_warp_node:
             time_warp_dict = {
                 "Class": time_warp_node.Class(),


client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -428,11 +435,3 @@ def _get_collected_frames(self, instance):
             collected_frames,
             first_frame
         )
-
-def _find_downstream_time_warp_node(start_node):
-    # HACK: no idea why calling `dependentNodes` the first time
-    # seems to always return nothing.
-    nuke.dependentNodes(nuke.INPUTS, [start_node])
```
BigRoy, line 435: Do we need to confirm whether this workaround is still needed?

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -187,8 +187,15 @@ def _set_additional_instance_data(
...
-            instance.data["transientData"]["node"]
+        time_warp_node = next(
+            (
+                dependent_node
+                for dependent_node in instance.data["transientData"][
+                    "node"
+                ].dependent(nuke.INPUTS, forceEvaluate=False)
+                if dependent_node.Class() == "TimeWarp"
+            ),
+            None,
```
BigRoy: Can we move it back so it still calls `_find_downstream_time_warp_node`? It's much clearer what the intent is than this `next()` with the generator, etc.

BigRoy:
@jakubjezek001 please confirm whether it's ok in practice to disable `evaluateAll` for these cases is fine - and also whether we indeed can avoid calling it twice or not.

client/ayon_nuke/plugins/publish/collect_writes.py
JoseCovaOS: Sure, have pushed it.

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -190,6 +190,7 @@ def _set_additional_instance_data(
         time_warp_node = _find_downstream_time_warp_node(
             instance.data["transientData"]["node"]
         )
+
```
BigRoy:
```suggestion
```

client/ayon_nuke/plugins/publish/collect_writes.py
```diff
@@ -430,9 +431,6 @@ def _get_collected_frames(self, instance):
...
 
 def _find_downstream_time_warp_node(start_node):
-    # HACK: no idea why calling `dependentNodes` the first time
-    # seems to always return nothing.
-    nuke.dependentNodes(nuke.INPUTS, [start_node])
-    for node in nuke.dependentNodes(nuke.INPUTS, [start_node]):
+    for node in start_node.dependent(nuke.INPUTS, forceEvaluate=False):
         if node.Class() == "TimeWarp":
-            return node
+            return node
```
BigRoy:
```suggestion
            return node

```

jakubjezek001 (approved):
yes this seems much better approach

client/ayon_nuke/plugins/publish/collect_writes.py
rdelillo: I think it's fine. `nuke.dependentNodes` actually calls `nuke.Node.dependent` under the hood:
https://learn.foundry.com/nuke/developers/17.0/pythondevguide/_modules/nuke_internal/overrides.html#dependentNodes

The fix really comes from `forceEvaluate=False` here. 
Foundry still warns: 
```text
  evaluateAll is an optional boolean defaulting to True. When this parameter is true, it forces a re-evaluation of the entire tree.
  This can be expensive, but otherwise could give incorrect results if nodes are expression-linked.
```
All thing considered, I'd say an incorrect result for expression-linked node is better than a crash/freeze.

rdelillo (approved):

client/ayon_nuke/plugins/publish/collect_writes.py
BigRoy: @rdelillo note that the **workaround** I was referring to was that apparently previously we called `dependentNodes` **twice** intentionally to avoid some bug because the first time around it _may_ return nothing (according to the comment that was also being removed with this PR).

rdelillo: > @rdelillo note that the **workaround** I was referring to was that apparently previously we called `dependentNodes` **twice** intentionally to avoid some bug because the first time around it _may_ return nothing (according to the comment that was also being removed with this PR).

I know, and I cannot reproduce it with Nuke 16.0/17.0.  I'm wondering if the bug might have been caused by the evaluation of the graph handled in a separate C++ process.

BigRoy (approved):

PR #339, opened 2026-07-23
feat: Collect loaded scene containers for upstream publish trace

jrsndl:
Introduces a Nuke Pyblish collector that records all currently loaded AYON containers in the scene as potential inputs for published instances. This creates an "upstream trace" by storing the representation IDs and user-friendly names of these containers on the publish instance data.

## Changelog Description
Adds inputs collector similar to Maya, Houdini and Fusion

The additional human readable input names are meant to be used for burnin (will require PR for core extract_burnin plugin).

### The naming:

```<project name>-<folder name>-product name-v version name```
* project name is added only if not matching current project (library)
* folder name is added only if not matching the current folder

### Examples:
MyLibrary-asset-renderLgtBeauty-v003
renderLgtCrypto-v001

## Possible Additional TODO
* settings for filtering only some products to be included in the user-friendly names
* settings to only include Nuke nodes (containers) that are connected to other Nuke nodes (some artists leave old containers loaded even if unused)

### DIFF
--- a/client/ayon_nuke/plugins/publish/collect_inputs.py
+++ b/client/ayon_nuke/plugins/publish/collect_inputs.py
@@ -1,0 +1,60 @@
+from collections import deque
+
+
+import pyblish.api
+from ayon_core.pipeline import (
+    registered_host,
+    get_current_context
+)
+
+
+class CollectUpstreamInputs(pyblish.api.InstancePlugin):
+    """Collect source input containers used for this publish.
+
+    This will include `inputs` data of which loaded publishes were used in the
+    generation of this publish. This leaves an upstream trace to what was used
+    as input.
+
+    """
+
+    label = "Collect Inputs"
+    order = pyblish.api.CollectorOrder + 0.2
+    hosts = ["nuke", "nukeassist"]
+    families = ["render", "image", "model", "source", "camera"]
+
+    def process(self, instance):
+
+        context = get_current_context()
+        project_name = context["project_name"]
+        folder_path = context["folder_path"]
+        folder_name = folder_path.split("/")[-1]
+
+        cache_key = "__cache_containers"
+        pretty_key = "inputRepresentationNames"
+        scene_containers = instance.context.data.get(cache_key, None)
+        if scene_containers is None:
+            # Query the scenes' containers if there's no cache yet
+            host = registered_host()
+            scene_containers = list(host.ls())
+            for container in scene_containers:
+                pretty_name = []
+                _project = container.get("project_name", "")
+                if _project != project_name:
+                   pretty_name.append(_project)
+                _namespace = container.get("namespace", "")
+                if _namespace != folder_name:
+                    pretty_name.append(_namespace)
+                pretty_name.append(container.get("name", ""))
+                pretty_name.append("v" + container.get("version", ""))
+                container["pretty_name"] = "-".join(pretty_name)
+            instance.context.data[cache_key] = scene_containers
+
+        if scene_containers:
+            inputs = [c["representation"] for c in scene_containers]
+            instance.data["inputRepresentations"] = inputs
+            self.log.debug("Collected inputs: %s" % inputs)
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by BigRoy — CHANGES_REQUESTED** [review]
There's a few issues here.

It's an InstancePlugin, but really wants to run over context to set the pretty names on the context (no idea why?). Also the pretty names logic uses e.g. `container["version"]` which Nuke does imprint, but which we may remove sometime soon to match with other hosts where we write only the limited data that we actually need to write out.

---

Also, it says collect upstream inputs, I do not see at all where this traces upstream to which nodes were used for this instance.

Consider a graph like:
```
A -> B -> C

D -> E -> F
```
With two separate streams.
When publishing F, it's not using A and hence I'd expect only D and E as input for F, not A, B or C.

---

I'd say:
- [ ] remove the pretty name stuff - if you want that, you'd likely use a different custom plug-in which we most likely NOT merge into ayon-nuke by default.
- [ ] focus on just the instance.data key `inputRepresentatons`
- [ ] collect only those for the instance that are actually 'upstream' for this output node.

**Comment by jrsndl** [comment]
OK, clearly I do not understand enough to do what needs to be done.
Apologies for sending PR way too early.

The use case is comp supe wants to know which cg render version(s) were used to produce the comp.
It can't answer which animation versions were used to produce the cg renders that produced the comp for example.

**Comment by BigRoy** [comment]
> The use case is comp supe wants to know which cg render version(s) were used to produce the comp.

This data is already collected in [core in `CollectSceneLoadedVersions`](https://github.com/ynput/ayon-core/blob/626494e30b508795ac0437b6e35a82e0a4cdfa70/client/ayon_core/plugins/publish/collect_scene_loaded_versions.py#L16), for [each workfile publish.](https://github.com/ynput/ayon-core/blob/12c7a8034a05ec443d4cd8a26f111153a52c3423/client/ayon_core/plugins/publish/integrate_inputlinks.py#L182-L226)

Some DCCs, like [Maya](https://github.com/ynput/ayon-maya/blob/06aa2445b4df511225b6795b283dd974040c0b5f/client/ayon_maya/plugins/publish/collect_inputs.py#L162-L161), [Houdini ](https://github.com/ynput/ayon-houdini/blob/18ebb3711c3f88fced4259aeae7759bc08490767/client/ayon_houdini/plugins/publish/collect_inputs.py#L102)and [Fusion ](https://github.com/ynput/ayon-fusion/blob/4bfaf0a943cb4aca178a42aa06e2ec51a63ff82d/client/ayon_fusion/plugins/publish/collect_inputs.py#L82)go more granular and collect it _per instance_ to see what's in that actual output filtering out what's _unused_ that was loaded within the same file. So that in e.g. in Maya when having two renderlayers, and character Hero is in layer A and character Villain in layer B that only Hero is connected to A, etc.

So from a publish in Maya it would currently store links:

```
workfile: Hero + Villain
A: Hero
B: Villain
```

This is already the case. In Nuke I think we only have the "workfile: all loaded containers" logic currently because that goes through core.

Note that each published version is also linked to the workfile.

So you can always query from the version if it's not a workfile, to get the relevant workfile (if it was published along) to get everything that was loaded in the scenefile.

Here's an example of what you can technically do with that data:
- [In AYON pipeline print a message if e.g. loaded look product was generated using a different model version than you have in your current scene.
](https://gist.github.com/BigRoy/eb196c9d4d55752687089a73d140a2bb)

---

I believe once https://github.com/ynput/ayon-frontend/issues/2170 gets implemented we may actually get some more visibility on this data on the web frontend too. Currently it's mostly being used/abused by studios with custom tools here and there.'

(Note that it links the versions, not the representations!)

jrsndl:
Thank you so much for the detailed explanation @BigRoy .
Amazing as always

BigRoy:
No worries, I'd still love to see the more granular one per instance in Nuke though ;) so feel free to match what e.g. Fusion does implementation wise. But my guess is that the data you needed actually lives somewhere already.

PR #344, opened 2026-08-03
[YN-0771]: Adds output type selection for intermediate presets

### LINKED ISSUE #343: YN-0771: Intermediate presets with more robust config

Currently:
- ProRes settings are hardcoded.
- H264 settings have limited access/customisation.

---

#

## Requested Enhancement
We need full control over the MOV render configuration, including:

Codec selection
Resolution settings
Output customisation
Render presets/configuration access
Ability to add additional MOV outputs if required
 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1506521293673992315)

_This issue was automatically created from AYON ticket [YN-0771]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d77a263b20fc4e6b89da97b1f1ea7e91)

jakubjezek001:
Introduces an `output_type` field on `IntermediateOutputModel` to choose between file-extension or custom write knobs. Includes a validator that enforces unique names on the knob list and reorganizes fields into their own sections.

## Changelog Description
Adds output type selection (extension vs custom write knobs) for intermediate presets. The new `output_type` field determines how output is written, with a fallback to "Defined by extension". Includes a validator ensuring unique names in the custom write knobs list and a reformat-nodes config that stores its data as a subfolder of representation files.

The changes focus on:

- **New `output_type` field** — A new enum-like resolver with two preset values ("Defined by extension", "Defined by custom write knobs") is added to the model, defaulting to extension. The corresponding data in the reformat nodes config is stored as a subfolder of representation files and written via `CustomWriteKnobsManager` for proper serialization.

- **Validation** — A validator on the `custom_write_knobs` list ensures unique names are preserved at runtime.

- **UI organization** — Related fields (output_type, custom_write_knobs) live under "Output definition" and add_custom_tags/extension/fill_missing_frames stay under "Representation definition".

## Testing notes:
1. Open intermediate preset editor and verify the new "Output type" dropdown appears
2. Test creating an intermediate with output_type set to custom write knobs and confirm `CustomWriteKnobsManager` serialization works correctly
3. Test validation by adding duplicate names in the knob list
4. Verify reformat-nodes data is written as subfolder of representation files

## Dependency
Close [ynput/ayon-nuke#343](https://github.com/ynput/ayon-nuke/issues/343)

## Related support tickets
[YN-0771](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d77a263b20fc4e6b89da97b1f1ea7e91)

### DIFF
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1183,7 +1183,7 @@ class ExporterReviewMov(ExporterReview):
                  klass,
                  instance,
                  name=None,
-                 ext=None,
+                 settings=None,
                  multiple_presets=True
                  ):
         # initialize parent class
@@ -1197,9 +1197,21 @@ class ExporterReviewMov(ExporterReview):
         self.write_colorspace = instance.data["colorspace"]
         self.color_channels = instance.data["color_channels"]
         self.formatting_data = instance.data["anatomyData"]
+        self.settings = settings or {}
 
         self.name = name or "baked"
-        self.ext = ext or "mov"
+        if settings["output_type"] == "extension":
+            self.ext = settings["extension"] or "mov"
+
+        elif settings["output_type"] == "custom_write_knobs":
+            for knob in settings["custom_write_knobs"]:
+                if knob["name"] == "file_type":
+                    self.ext = knob["value"]
+                    break
+            else:
+                self.ext = "mov"
+        else:
+            self.ext = "mov"
 
         # set frame start / end and file name to self
         self.get_file_info()
@@ -1409,28 +1421,43 @@ class ExporterReviewMov(ExporterReview):
         self.log.debug(f"Path: {self.path}")
 
         write_node["file"].setValue(str(self.path))
-        write_node["file_type"].setValue(str(self.ext))
-        write_node["channels"].setValue(str(self.color_channels))
 
-        # Knobs `meta_codec` and `mov64_codec` are not available on centos.
-        # TODO shouldn't this come from settings on outputs?
-        try:
-            write_node["meta_codec"].setValue("ap4h")
-        except Exception:
-            self.log.info("`meta_codec` knob was not found")
+        if self.settings["output_type"] == "extension":
+            write_node["file_type"].setValue(str(self.ext))
+            write_node["channels"].setValue(str(self.color_channels))
 
-        try:
-            write_node["mov64_codec"].setValue("ap4h")
-            write_node["mov64_fps"].setValue(float(fps))
-        except Exception:
-            self.log.info("`mov64_codec` knob was not found")
+            # Knobs `meta_codec` and `mov64_codec` are not available on centos.
+            # TODO shouldn't this come from settings on outputs?
+            try:
+                write_node["meta_codec"].setValue("ap4h")
+            except Exception:
+                self.log.info("`meta_codec` knob was not found")
 
-        try:
-            write_node["mov64_write_timecode"].setValue(1)
-        except Exception:
-            self.log.info("`mov64_write_timecode` knob was not found")
+            try:
+                write_node["mov64_codec"].setValue("ap4h")
+                write_node["mov64_fps"].setValue(float(fps))
+            except Exception:
+                self.log.info("`mov64_codec` knob was not found")
 
-        write_node["raw"].setValue(1)
+            try:
+                write_node["mov64_write_timecode"].setValue(1)
+            except Exception:
+                self.log.info("`mov64_write_timecode` knob was not found")
+
+            write_node["raw"].setValue(1)
+
+        elif self.settings["output_type"] == "custom_write_knobs":
+            write_node["file_type"].setValue(str(self.ext))
+            write_node["channels"].setValue(str(self.color_channels))
+            write_node["raw"].setValue(1)
+
+            for knob in self.settings["custom_write_knobs"]:
+                try:
+                    write_node[knob["name"]].setValue(knob["value"])
+                except Exception:
+                    self.log.info(
+                        f"`{knob['name']}` knob was not found on Write node"
+                    )
 
         # connect
         write_node.setInput(0, self.previous_node)
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1205,14 +1205,23 @@ class ExporterReviewMov(ExporterReview):
 
         output_type = self._get_output_type()
         if output_type == "custom_write_knobs":
+            file_type = None
             for knob in self.settings.get("custom_write_knobs") or []:
-                if knob["name"] == "file_type":
-                    self.ext = knob["text"] or "mov"
+                if knob.get("name") == "file_type":
+                    file_type = knob.get("text")
                     break
-            else:
-                self.ext = "mov"
+
+            if not file_type:
+                raise ValueError(
+                    "Invalid custom_write_knobs settings: "
+                    "expected a non-empty 'file_type' "
+                    "knob text value when output_type is "
+                    "'custom_write_knobs'."
+                )
+
+            self.ext = file_type
         else:
-            self.ext = self.settings.get("extension") or "mov"
+            self.ext = self.settings.get("extension", "mov")
 
         # set frame start / end and file name to self
         self.get_file_info()
@@ -1451,7 +1460,10 @@ class ExporterReviewMov(ExporterReview):
             if "mov64_fps" in write_node.knobs():
                 write_node["mov64_fps"].setValue(float(fps))
         else:
-            self._configure_write_node(write_node, fps)
+            raise ValueError(
+                f"Invalid output type: {output_type}. "
+                "Expected 'extension' or 'custom_write_knobs'."
+            )
 
         # connect
         write_node.setInput(0, self.previous_node)
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1185,7 +1185,7 @@ class ExporterReviewMov(ExporterReview):
                  klass,
                  instance,
                  name=None,
-                 settings=None,
+                 write_knobs=None,
                  multiple_presets=True
                  ):
         # initialize parent class
@@ -1199,30 +1199,9 @@ class ExporterReviewMov(ExporterReview):
         self.write_colorspace = instance.data["colorspace"]
         self.color_channels = instance.data["color_channels"]
         self.formatting_data = instance.data["anatomyData"]
-        self.settings = settings or {}
+        self.custom_write_knobs = write_knobs["custom"]
         self.name = name or "baked"
-
-        custom_write_knobs = self.settings.get("custom_write_knobs", [])
-        if custom_write_knobs:
-            file_type = None
-            for knob in custom_write_knobs:
-                if knob.get("name") == "file_type":
-                    file_type = knob.get("text")
-                    break
-
-            if not file_type:
-                raise ValueError(
-                    "Invalid custom_write_knobs settings: "
-                    "expected a non-empty 'file_type' "
-                    "knob text value when output_type is "
-                    "'custom_write_knobs'."
-                )
-
-            self.ext = file_type
-        else:
-            # backwards compatibility for old settings to
-            # get extension
-            self.ext = self.settings.get("extension", "mov")
+        self.ext = write_knobs["file_type"]
 
         # set frame start / end and file name to self
         self.get_file_info()
@@ -1431,10 +1410,8 @@ class ExporterReviewMov(ExporterReview):
         write_node = nuke.createNode("Write")
         self.log.debug(f"Path: {self.path}")
         write_node["file"].setValue(str(self.path))
-
-        custom_write_knobs = self.settings.get("custom_write_knobs") or []
-        if custom_write_knobs:
-            for knob in custom_write_knobs:
+        if self.custom_write_knobs:
+            for knob in self.custom_write_knobs:
                 if knob["name"] in {
                     "file", "file_type",
                     "channels", "raw", "mov64_fps"
@@ -1446,11 +1423,7 @@ class ExporterReviewMov(ExporterReview):
                 )
                 write_node[knob["name"]].setValue(value)
 
-            self._set_write_node_defaults(write_node, fps)
-        else:
-            # backward compatibility for old settings
-            # with only extension defined
-            self._set_write_node_defaults(write_node, fps)
+        self._set_write_node_defaults(write_node, fps)
 
         # connect
         write_node.setInput(0, self.previous_node)
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1479,6 +1479,7 @@ class ExporterReviewMov(ExporterReview):
         Args:
             write_node: The write node to configure.
             custom_knobs: List of custom knob configurations.
+            log: Logger for logging warnings and information.
         """
         if not custom_knobs:
             return
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1206,7 +1206,7 @@ class ExporterReviewMov(ExporterReview):
         elif settings["output_type"] == "custom_write_knobs":
             for knob in settings["custom_write_knobs"]:
                 if knob["name"] == "file_type":
-                    self.ext = knob["value"]
+                    self.ext = knob["text"] or "mov"
                     break
             else:
                 self.ext = "mov"
@@ -1450,10 +1450,24 @@ class ExporterReviewMov(ExporterReview):
             write_node["file_type"].setValue(str(self.ext))
             write_node["channels"].setValue(str(self.color_channels))
             write_node["raw"].setValue(1)
+            if "mov64_fps" in write_node.knobs():
+                write_node["mov64_fps"].setValue(float(fps))
 
             for knob in self.settings["custom_write_knobs"]:
                 try:
-                    write_node[knob["name"]].setValue(knob["value"])
+                    if knob["type"] == "text":
+                        write_node[knob["name"]].setValue(str(knob["text"]))
+                    elif knob["type"] == "number":
+                        write_node[knob["name"]].setValue(int(knob["number"]))
+                    elif knob["type"] == "decimal_number":
+                        write_node[knob["name"]].setValue(float(knob["decimal_number"]))
+                    elif knob["type"] == "boolean":
+                        write_node[knob["name"]].setValue(bool(knob["boolean"]))
+                    else:
+                        self.log.warning(
+                            f"Knob type `{knob['type']}` is not supported"
+                        )
+
                 except Exception:
                     self.log.info(
                         f"`{knob['name']}` knob was not found on Write node"
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1203,10 +1203,10 @@ class ExporterReviewMov(ExporterReview):
         self.settings = settings or {}
         self.name = name or "baked"
 
-        output_type = self._get_output_type()
-        if output_type == "custom_write_knobs":
+        custom_write_knobs = self.settings.get("custom_write_knobs", [])
+        if custom_write_knobs:
             file_type = None
-            for knob in self.settings.get("custom_write_knobs") or []:
+            for knob in custom_write_knobs:
                 if knob.get("name") == "file_type":
                     file_type = knob.get("text")
                     break
@@ -1221,6 +1221,8 @@ class ExporterReviewMov(ExporterReview):
 
             self.ext = file_type
         else:
+            # backwards compatibility for old settings to
+            # get extension
             self.ext = self.settings.get("extension", "mov")
 
         # set frame start / end and file name to self
@@ -1244,14 +1246,6 @@ class ExporterReviewMov(ExporterReview):
                 self.fhead, self.name, after_head, self.ext)
         self.path = os.path.join(
             self.staging_dir, self.file).replace("\\", "/")
-
-    def _get_output_type(self) -> Literal["extension", "custom_write_knobs"]:
-        """Get the output type from settings.
-
-        Returns:
-            Literal["extension", "custom_write_knobs"]: output type
-        """
-        return self.settings.get("output_type", "extension")
 
     def clean_nodes(self, node_name):
         for node in self._temp_nodes[node_name]:
@@ -1439,12 +1433,9 @@ class ExporterReviewMov(ExporterReview):
         self.log.debug(f"Path: {self.path}")
         write_node["file"].setValue(str(self.path))
 
-        output_type = self._get_output_type()
-        if output_type == "extension":
-            self._set_write_node_defaults(write_node, fps)
-
-        elif output_type == "custom_write_knobs":
-            for knob in self.settings.get("custom_write_knobs") or []:
+        custom_write_knobs = self.settings.get("custom_write_knobs") or []
+        if custom_write_knobs:
+            for knob in custom_write_knobs:
                 if knob["name"] in {
                     "file", "file_type",
                     "channels", "raw", "mov64_fps"
@@ -1458,10 +1449,9 @@ class ExporterReviewMov(ExporterReview):
 
             self._set_write_node_defaults(write_node, fps)
         else:
-            raise ValueError(
-                f"Invalid output type: {output_type}. "
-                "Expected 'extension' or 'custom_write_knobs'."
-            )
+            # backward compatibility for old settings
+            # with only extension defined
+            self._set_write_node_defaults(write_node, fps)
 
         # connect
         write_node.setInput(0, self.previous_node)
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1198,20 +1198,18 @@ class ExporterReviewMov(ExporterReview):
         self.color_channels = instance.data["color_channels"]
         self.formatting_data = instance.data["anatomyData"]
         self.settings = settings or {}
-
         self.name = name or "baked"
-        if settings["output_type"] == "extension":
-            self.ext = settings["extension"] or "mov"
 
-        elif settings["output_type"] == "custom_write_knobs":
-            for knob in settings["custom_write_knobs"]:
+        output_type = self.settings.get("output_type", "extension")
+        if output_type == "custom_write_knobs":
+            for knob in self.settings.get("custom_write_knobs") or []:
                 if knob["name"] == "file_type":
                     self.ext = knob["text"] or "mov"
                     break
             else:
                 self.ext = "mov"
         else:
-            self.ext = "mov"
+            self.ext = self.settings.get("extension") or "mov"
 
         # set frame start / end and file name to self
         self.get_file_info()
@@ -1419,10 +1417,10 @@ class ExporterReviewMov(ExporterReview):
         # Write node
         write_node = nuke.createNode("Write")
         self.log.debug(f"Path: {self.path}")
-
         write_node["file"].setValue(str(self.path))
 
-        if self.settings["output_type"] == "extension":
+        output_type = self.settings.get("output_type", "extension")
+        if output_type == "extension":
             write_node["file_type"].setValue(str(self.ext))
             write_node["channels"].setValue(str(self.color_channels))
 
@@ -1446,14 +1444,16 @@ class ExporterReviewMov(ExporterReview):
 
             write_node["raw"].setValue(1)
 
-        elif self.settings["output_type"] == "custom_write_knobs":
+        elif output_type == "custom_write_knobs":
+            # force setting file type and channels
+            # to avoid export issues
             write_node["file_type"].setValue(str(self.ext))
             write_node["channels"].setValue(str(self.color_channels))
             write_node["raw"].setValue(1)
             if "mov64_fps" in write_node.knobs():
                 write_node["mov64_fps"].setValue(float(fps))
 
-            for knob in self.settings["custom_write_knobs"]:
+            for knob in self.settings.get("custom_write_knobs") or []:
                 try:
                     if knob["type"] == "text":
                         write_node[knob["name"]].setValue(str(knob["text"]))
@@ -1472,6 +1472,29 @@ class ExporterReviewMov(ExporterReview):
                     self.log.info(
                         f"`{knob['name']}` knob was not found on Write node"
                     )
+
+        else:
+            write_node["file_type"].setValue(str(self.ext))
+            write_node["channels"].setValue(str(self.color_channels))
+            write_node["raw"].setValue(1)
+
+            # Knobs `meta_codec` and `mov64_codec` are not available on centos.
+            # TODO shouldn't this come from settings on outputs?
+            try:
+                write_node["meta_codec"].setValue("ap4h")
+            except Exception:
+                self.log.info("`meta_codec` knob was not found")
+
+            try:
+                write_node["mov64_codec"].setValue("ap4h")
+                write_node["mov64_fps"].setValue(float(fps))
+            except Exception:
+                self.log.info("`mov64_codec` knob was not found")
+
+            try:
+                write_node["mov64_write_timecode"].setValue(1)
+            except Exception:
+                self.log.info("`mov64_write_timecode` knob was not found")
 
         # connect
         write_node.setInput(0, self.previous_node)
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1485,19 +1485,16 @@ class ExporterReviewMov(ExporterReview):
             return
 
         for knob in custom_knobs:
-            if knob["name"] in {
-                "file", "file_type",
-                "channels", "raw", "mov64_fps"
-            }:
-                log.warning(
-                    f"Skipping to set custom knob '{knob['name']}' "
-                    "as it is already set in defaults."
-                )
-                continue
             to_type = knob["type"]
             value = convert_knob_value_to_correct_type(
                 to_type, knob[to_type]
             )
+            if knob["name"] not in write_node.knobs():
+                log.warning(
+                    f"Knob '{knob['name']}' does not exist on the write node. "
+                    "Skipping setting this knob."
+                )
+                continue
             write_node[knob["name"]].setValue(value)
 
 
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1,4 +1,3 @@
-from typing import Literal
 
 import nuke
 import re
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1,3 +1,5 @@
+from typing import Literal
+
 import nuke
 import re
 import os
@@ -1201,7 +1203,7 @@ class ExporterReviewMov(ExporterReview):
         self.settings = settings or {}
         self.name = name or "baked"
 
-        output_type = self.settings.get("output_type", "extension")
+        output_type = self._get_output_type()
         if output_type == "custom_write_knobs":
             for knob in self.settings.get("custom_write_knobs") or []:
                 if knob["name"] == "file_type":
@@ -1233,6 +1235,14 @@ class ExporterReviewMov(ExporterReview):
                 self.fhead, self.name, after_head, self.ext)
         self.path = os.path.join(
             self.staging_dir, self.file).replace("\\", "/")
+
+    def _get_output_type(self) -> Literal["extension", "custom_write_knobs"]:
+        """Get the output type from settings.
+
+        Returns:
+            Literal["extension", "custom_write_knobs"]: output type
+        """
+        return self.settings.get("output_type", "extension")
 
     def clean_nodes(self, node_name):
         for node in self._temp_nodes[node_name]:
@@ -1420,7 +1430,7 @@ class ExporterReviewMov(ExporterReview):
         self.log.debug(f"Path: {self.path}")
         write_node["file"].setValue(str(self.path))
 
-        output_type = self.settings.get("output_type", "extension")
+        output_type = self._get_output_type()
         if output_type == "extension":
             self._configure_write_node(write_node, fps)
 
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -45,6 +45,7 @@ from .lib import (
     get_work_default_directory,
     link_knobs,
     get_version_from_path,
+    convert_knob_value_to_correct_type,
 )
 from .pipeline import (
     list_instances,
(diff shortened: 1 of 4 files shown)


client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1197,21 @@ def __init__(self,
         self.write_colorspace = instance.data["colorspace"]
         self.color_channels = instance.data["color_channels"]
         self.formatting_data = instance.data["anatomyData"]
+        self.settings = settings or {}
 
         self.name = name or "baked"
-        self.ext = ext or "mov"
+        if settings["output_type"] == "extension":
+            self.ext = settings["extension"] or "mov"
```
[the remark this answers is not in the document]
moonyuet: [21810f2](https://github.com/ynput/ayon-nuke/pull/344/commits/21810f252a2134147fb8711b14d1e111d4f9c6d1)

server/settings/publish_plugins.py
```diff
@@ -191,6 +218,12 @@ class IntermediateOutputModel(BaseSettingsModel):
         enum_resolver=_handle_missing_frames_enum
     )
```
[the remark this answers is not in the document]
moonyuet: [74226c2](https://github.com/ynput/ayon-nuke/pull/344/commits/74226c26d2fb76feca5cea52a4d0001cb2171870)

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1417,84 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
+                    elif knob["type"] == "number":
+                        write_node[knob["name"]].setValue(int(knob["number"]))
+                    elif knob["type"] == "decimal_number":
+                        write_node[knob["name"]].setValue(float(knob["decimal_number"]))
+                    elif knob["type"] == "boolean":
+                        write_node[knob["name"]].setValue(bool(knob["boolean"]))
+                    else:
+                        self.log.warning(
+                            f"Knob type `{knob['type']}` is not supported"
+                        )
```
jakubjezek001: Can't we just use some of nuke/api/lib function for this? (convert_knob_value_to_correct_type, set_node_knobs_from_settings)

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1417,84 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
+            write_node["raw"].setValue(1)
+
+        elif output_type == "custom_write_knobs":
+            # force setting file type and channels
+            # to avoid export issues
+            write_node["file_type"].setValue(str(self.ext))
+            write_node["channels"].setValue(str(self.color_channels))
+            write_node["raw"].setValue(1)
+            if "mov64_fps" in write_node.knobs():
+                write_node["mov64_fps"].setValue(float(fps))
```
jakubjezek001: I would rather set most of those after the custom write knobs are applied. or perhaps make sure there is a list of knob names we are automatically excluding from 'custom_write_knobs', just in case someone is adding them into settings. Also please lets log them as info to let know user that those are ignored from settings but rather hardcoded as required by the workflow.

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1417,84 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
+
+                except Exception:
+                    self.log.info(
+                        f"`{knob['name']}` knob was not found on Write node"
+                    )
+
+        else:
+            write_node["file_type"].setValue(str(self.ext))
+            write_node["channels"].setValue(str(self.color_channels))
+            write_node["raw"].setValue(1)
```
jakubjezek001: this is repeating three times, so we could perhaps set them only once for all, right?

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1417,84 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
+            try:
+                write_node["mov64_codec"].setValue("ap4h")
+                write_node["mov64_fps"].setValue(float(fps))
+            except Exception:
+                self.log.info("`mov64_codec` knob was not found")
+
+            try:
+                write_node["mov64_write_timecode"].setValue(1)
+            except Exception:
+                self.log.info("`mov64_write_timecode` knob was not found")
```
jakubjezek001: could we also abstract this section to be reused rather then duplicate the code?

server/settings/publish_plugins.py
```diff
@@ -176,21 +188,45 @@ class IntermediateOutputModel(BaseSettingsModel):
...
+    custom_write_knobs: list[KnobModel] = SettingsField(
+        default_factory=list,
+        title="Custom Write Knobs",
+        section="Output definition",
+    )
     extension: str = SettingsField(
         "mov",
-        title="File extension"
+        title="File extension",
+        section="Representation definition",
```
jakubjezek001:
```suggestion
```
no need to add this line. This is field which is part of `Output definition` and that had been already added at `Output type` field

jakubjezek001:
anyway I was testing the code on some example use cases and all was working as expected. Once the code is cleared it will be Approvable.

client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1197,19 @@ def __init__(self,
         self.write_colorspace = instance.data["colorspace"]
         self.color_channels = instance.data["color_channels"]
         self.formatting_data = instance.data["anatomyData"]
-
+        self.settings = settings or {}
         self.name = name or "baked"
-        self.ext = ext or "mov"
+
+        output_type = self.settings.get("output_type", "extension")
```
BigRoy: Can we get a type hint on what values `output_type` is supposed to be, e.g. `Literal["extension", "custom_write_knobs"]`

BigRoy: Or maybe add a method `def _get_output_type(self) -> Literal["extension", "custom_write_knobs"]` just so that wherever we call this we can rely on the resulting type?

server/settings/publish_plugins.py
```diff
@@ -176,21 +188,44 @@ class IntermediateOutputModel(BaseSettingsModel):
...
+        default="extension",
+        enum_resolver=_intermediate_preset_output_types,
+        conditional_enum=True,
+        section="Output definition",
+    )
+    custom_write_knobs: list[KnobModel] = SettingsField(
+        default_factory=list,
+        title="Custom Write Knobs",
+        section="Output definition",
+    )
```
BigRoy: Can we add a description of what this does and what it should be used for. What custom write knobs? Where? what are we affecting? As a Nuke lead - how do I know what to put here and when to use it?

server/settings/publish_plugins.py
```diff
@@ -176,21 +188,44 @@ class IntermediateOutputModel(BaseSettingsModel):
...
+        title="Output type",
+        default="extension",
+        enum_resolver=_intermediate_preset_output_types,
+        conditional_enum=True,
+        section="Output definition",
+    )
+    custom_write_knobs: list[KnobModel] = SettingsField(
+        default_factory=list,
+        title="Custom Write Knobs",
+        section="Output definition",
```
BigRoy: This doesn't need the section, right? it's always underneath `output_type`?

client/ayon_nuke/api/plugin.py
moonyuet: [1408d88](https://github.com/ynput/ayon-nuke/pull/344/commits/1408d88092978c699db77154ddc11d0e68b532bd)

server/settings/publish_plugins.py
moonyuet: [b14f88e](https://github.com/ynput/ayon-nuke/pull/344/commits/b14f88e2d4006fcc236bd48ccc33c4d9394ccdb2)

server/settings/publish_plugins.py
moonyuet: [b14f88e](https://github.com/ynput/ayon-nuke/pull/344/commits/b14f88e2d4006fcc236bd48ccc33c4d9394ccdb2)

client/ayon_nuke/api/plugin.py
moonyuet: [5a37de5](https://github.com/ynput/ayon-nuke/pull/344/commits/5a37de5fa6e361bbc3d05ef27ca67271b32f70aa)

client/ayon_nuke/api/plugin.py
moonyuet: [5a37de5](https://github.com/ynput/ayon-nuke/pull/344/commits/5a37de5fa6e361bbc3d05ef27ca67271b32f70aa)

client/ayon_nuke/api/plugin.py
moonyuet: [5a37de5](https://github.com/ynput/ayon-nuke/pull/344/commits/5a37de5fa6e361bbc3d05ef27ca67271b32f70aa)

client/ayon_nuke/api/plugin.py
moonyuet: [5a37de5](https://github.com/ynput/ayon-nuke/pull/344/commits/5a37de5fa6e361bbc3d05ef27ca67271b32f70aa)

moonyuet:
@jakubjezek001 @BigRoy can you review again? Thanks!

client/ayon_nuke/plugins/publish/extract_review_intermediates.py
```diff
@@ -111,7 +111,7 @@ def process(self, instance):
 
                 # create exporter instance
                 exporter = plugin.ExporterReviewMov(
-                    self, instance, o_name, o_data["extension"],
+                    self, instance, o_name, o_data,
```
rdelillo: I'd explicit the kwargs here.
```suggestion
                    self,
                    instance,
                    name=o_name,
                    settings=o_data,
                    multiple_presets=multiple_presets,
```

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1428,28 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
-            self.log.info("`mov64_write_timecode` knob was not found")
+                value = convert_knob_value_to_correct_type(write_node, knob)
+                write_node[knob["name"]].setValue(value)
 
-        write_node["raw"].setValue(1)
+            self._set_write_node_defaults(write_node)
+            if "mov64_fps" in write_node.knobs():
+                write_node["mov64_fps"].setValue(float(fps))
+        else:
+            self._configure_write_node(write_node, fps)
```
rdelillo: is it supported that `output_type not in ("extension", "custom_write_knobs")` after `output_type = self._get_output_type()`.
Maybe I'm wrong but I'd have expected a ValueError here.

rdelillo (commented):
Confirm current code works and render intermediate as expected !

@jakubjezek001 I'm not sure the need for `extension` option here.
E.g. I tried to set `extension = mp4` but still got the following issue `the selected codec is not supported in this container` cause it was defaulting to prores.

Shouldn't it all done through `custom_write_knobs` option ? That sounds more explicit and future-proof to me.

client/ayon_nuke/api/plugin.py
```diff
@@ -1479,6 +1500,43 @@ def _connect_to_above_nodes(self, node, product_name, message):
...
+        """Set write node configuration.
+
+        Args:
+            write_node: The write node to configure.
+            fps: Frames per second for the write node.
+        """
+        self._set_write_node_defaults(write_node)
+
+        # Knobs `meta_codec` and `mov64_codec` are not available on centos.
+        # TODO shouldn't this come from settings on outputs?
```
rdelillo: I'm not sure I understand this,
The write node get initialized from path.
Then this function enforces ProRes + timecode when possible (basically when a mov/mp4 extension is set).

Is it just because that's how it used to be done in previous version?

I'd have thought this function should just leave the knob untouched if initialized from extension. And potentially we need to change the default server settings to be `"output_type": "custom_write_knobs"` and preserve backward compatibility ?

client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1200,19 @@ def __init__(self,
...
         self.formatting_data = instance.data["anatomyData"]
-
+        self.settings = settings or {}
         self.name = name or "baked"
-        self.ext = ext or "mov"
+
+        output_type = self._get_output_type()
+        if output_type == "custom_write_knobs":
+            for knob in self.settings.get("custom_write_knobs") or []:
+                if knob["name"] == "file_type":
```
BigRoy: If the file type is in the custom write knobs - shouldn't it have a sensible value by the design of it (e.g. requiring non-empty values in settings?). Seems a bit odd to fall back to `mov`?

client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1200,19 @@ def __init__(self,
...
+                    self.ext = knob["text"] or "mov"
+                    break
+            else:
+                self.ext = "mov"
+        else:
+            self.ext = self.settings.get("extension") or "mov"
```
BigRoy: Not too keen on the magical fallback to `mov`. I'd rather be explicit in settings, right?

client/ayon_nuke/api/plugin.py
BigRoy: Agreed - this else clause will just hide issues or bugs.

BigRoy (commented):
I agree with @rdelillo that we seem to mix old fallbacks with the new design.

client/ayon_nuke/api/plugin.py
moonyuet: Agree to a large extent - As we have the custom write knobs settings , we can just leave the settings untouched as they are already user-defined settings. The only thing I would keep is the fps settings so that it can be synced with the fps set within the project.

client/ayon_nuke/api/plugin.py
moonyuet: [7dc4556](https://github.com/ynput/ayon-nuke/pull/344/commits/7dc455609bafbe33d2a558c7b8bf081483de6a95)

client/ayon_nuke/api/plugin.py
moonyuet: [7dc4556](https://github.com/ynput/ayon-nuke/pull/344/commits/7dc455609bafbe33d2a558c7b8bf081483de6a95)

client/ayon_nuke/api/plugin.py
moonyuet: [7dc4556](https://github.com/ynput/ayon-nuke/pull/344/commits/7dc455609bafbe33d2a558c7b8bf081483de6a95)

client/ayon_nuke/api/plugin.py
moonyuet: [d7f4750](https://github.com/ynput/ayon-nuke/pull/344/commits/d7f47506188250d7350bb97d63a6afe21f94b8d7) and [7a3212f](https://github.com/ynput/ayon-nuke/pull/344/commits/7a3212f16917e637ad4eb6ad0bd6b911fee9d2a3)

rdelillo (approved):
Approving cause this code works as exected.

As mentioned, I'd have one side question what's the benefit of the `extension` option here ?
Is it a UX/UI simplification ? `custom_write_knobs` option let you set everything directly including a mandatory `file_type` knob, isn't that enough to accomodate all use-case ?

moonyuet:
> As mentioned, I'd have one side question what's the benefit of the `extension` option here ? Is it a UX/UI simplification ? `custom_write_knobs` option let you set everything directly including a mandatory `file_type` knob, isn't that enough to accomodate all use-case ?

I actually agree that we can just have the `custom_write_knobs` and we can have the backward compatibility for that. @BigRoy @jakubjezek001 thought?

BigRoy:
Most important to me is that a studio admin or cg supe/lead that might be responsible for configuring these settings has an easy time to actually configure what he wants here. If it's hard to know which 20 things to define, the best we can do is lay that out very clearly.

1. If it's complex to configure, we're doing things wrong.
2. If there's twenty ways to do the exact same thing - I think we're also doing something wrong.

So preferably we find what is the most understandable for that CG supe/lead who knows a bit of Nuke, can configure it quickly and is still flexible enough to configure their write exactly the way they want in the majority of cases. If e.g. file type is absolutely mandatory (it can never work without it) then I'm fine with that being an explicit setting that lives close to it just to make things easier for configuring things.

But since I lack Nuke expertise, I'll let you decide what feels most comfortable from a Nuke lead perspective.

jakubjezek001:
Very good point @rdelillo . We could perhaps really just remove the switch between **extension** and **custom_write_knobs**, and instead to keep it always as **output_write_knobs** with good-backward compatible default. This way we would made all the implicit _prores codec_ conditioned code explicit from the settings. 

For this solution which is definitely preferable we will need to add settings conversion methods, Issue here is that some productions might be using extension set to **png**, **exr**, **jpg**, **mp4**, **mov**. Here I would perhaps make sure each of those extension is covered with some dissent large data quality. 

> But since I lack Nuke expertise...

Are you really still lacking Nuke experiences @BigRoy ? :D

moonyuet:
@BigRoy @rdelillo can you review again?

client/ayon_nuke/api/plugin.py
BigRoy, line 1180: Can we put these on the `__init__`  as docstring, add the missing arguments and also add `settings` and explain what it is. Like that it should represent a single entry from `ExtractReviewIntermediates.outputs` setting, etc. 

Also, does it make more sense now to actually only pass in `custom_write_knobs`?

client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1199,30 @@ def __init__(self,
...
+                    "expected a non-empty 'file_type' "
+                    "knob text value when output_type is "
+                    "'custom_write_knobs'."
+                )
+
+            self.ext = file_type
+        else:
+            # backwards compatibility for old settings to
+            # get extension
+            self.ext = self.settings.get("extension", "mov")
```
BigRoy: `extension` key has been removed from settings, so this fallback makes no more sense.

client/ayon_nuke/api/plugin.py
```diff
@@ -1197,9 +1199,30 @@ def __init__(self,
...
+                    break
+
+            if not file_type:
+                raise ValueError(
+                    "Invalid custom_write_knobs settings: "
+                    "expected a non-empty 'file_type' "
+                    "knob text value when output_type is "
+                    "'custom_write_knobs'."
+                )
+
```
BigRoy: If this is an **absolute** requirement I still wonder whether it makes sense to rely on this value existing in a list of settings where you CAN remove it. Should it not be the case that this then is the only entry that is explicitly a static setting? (That was my point as to what I mentioned before). Aren't we making the UX for the admin worse now in that I _can_ configure broken settings and need to know I need to add and maintain a `file_type` entry?

If we kept `extension`, made it clear that it's the `file_type` knob - then is it really pretty to force it into `custom_write_knobs`?

server/settings/publish_plugins.py
```diff
@@ -176,21 +176,38 @@ class IntermediateOutputModel(BaseSettingsModel):
...
+    )
 
     fill_missing_frames:str = SettingsField(
         title="Handle missing frames",
         default="0",
         description="What to do about missing frames from entity frame range."
                     "Used for filling gaps for Custom Frames",
-        enum_resolver=_handle_missing_frames_enum
+        enum_resolver=_handle_missing_frames_enum,
+        section="Representation definition",
```
BigRoy: We don't need to be repeating the section right? It only needs to be on the first of these?

server/settings/publish_plugins.py
```diff
@@ -176,21 +176,38 @@ class IntermediateOutputModel(BaseSettingsModel):
     reformat_nodes_config: ReformatNodesConfigModel = SettingsField(
         default_factory=ReformatNodesConfigModel,
         title="Reformat Nodes")
-    extension: str = SettingsField(
-        "mov",
-        title="File extension"
```
BigRoy, line 181: Are we missing settings conversions still for this? (or is GH just not showing it for me? 🤔 )

BigRoy: But I think first we should really consider: https://github.com/ynput/ayon-nuke/pull/344#discussion_r3727428134 and https://github.com/ynput/ayon-nuke/pull/344#issuecomment-5201458881

@jakubjezek001 @rdelillo 

If it must be set, we want to make it very hard for someone to misconfigure it - at best, impossible. If this key is required, a dedicated setting for the fact that it's required where you cannot accidentally remove it is the best approach.

moonyuet: > Are we missing settings conversions still for this? (or is GH just not showing it for me? 🤔 )

This is indeed missing. But as soon as it removed. what should we do to implement it back?

BigRoy: Let's first finish the discussion - my guess is we'll end of keeping "Extension" (file type) separate but that's my guess. Might make settings conversion redundant here to begin with.

client/ayon_nuke/api/plugin.py
moonyuet: For the future yes, but for backward compatibility, it is just a bit difficult as we need to also take care of "extension" from the setting. It depends on whether we would like to support the changes for development mode.

client/ayon_nuke/api/plugin.py
moonyuet: > If this is an **absolute** requirement I still wonder whether it makes sense to rely on this value existing in a list of settings where you CAN remove it. Should it not be the case that this then is the only entry that is explicitly a static setting? (That was my point as to what I mentioned before). Aren't we making the UX for the admin worse now in that I _can_ configure broken settings and need to know I need to add and maintain a `file_type` entry?
> 
> If we kept `extension`, made it clear that it's the `file_type` knob - then is it really pretty to force it into `custom_write_knobs`?

This exception is raised here to just in case users accidentally remove the value from the addon settings. 
But as we have designed a setting, we assume the users do have ability to know how to define the settings.

server/settings/publish_plugins.py
jakubjezek001: We could add settings validation on the existing knob set and if `file_type` is missing then it would rise error. So it would mean **file_type** should be single minimal required knob name with defined value (without dot so '.mov' would be also failing). If we secure it this way then we will be fitting also backward compatibility. The settings conversion methods with solid default set of knobs per detected extension values is still relevant, if we decide to go this way.

jakubjezek001: The same validation for knob content could be perhaps used for knobs we want to exclude so user knows why they are being ignored (like **channels** and **raw** or **file**)

jakubjezek001: To be honest perhaps being able to alter `mov64_fps` for intermediates makes sense, I am facing the similar approach now in **batchdelivery**. But really we would need to add the `fps` key with the changed value to the representation data so it is also respected down the line. This might be of course suggesting next enhancement issue.

BigRoy: > We could add settings validation on the existing knob set and if file_type is missing then it would rise error. So it would mean file_type should be single minimal required knob name with defined value (without dot so '.mov' would be also failing). If we secure it this way then we will be fitting also backward compatibility. The settings conversion methods with solid default set of knobs per detected extension values is still relevant, if we decide to go this way.

We could do that - but as soon as, in any way, you get settings where it's not existing the settings model doesn't load and you're breaking your settings. So the settings conversion would need to be SUPER stable.

In that case:
- We must validate the type as well?
- I think we should drop the "Custom" part. It's just the Write Knobs to customize how the write node behaves.

And in that scenario. I'd actually do that:
- remove `file_type` from the write knobs list, add a validation that disallows you to add it!
- then above it, add a `file_type: str` setting so that you're enforced to set it.

So you'd get:

```py
# root
ExtractReviewIntermediates.outputs[x].write_knobs

# and the file type + custom knobs
write_knobs.file_type: str  # explicit file type knob setting
write_knobs.custom: list[KnobModel] # add your own extra knobs, disallow `file_type`
```

also, this setting is currently under the "Baking additional" setting.

image
 
Does it only apply in certain cases (if a certain checkbox is checked?) or always, because if it the baking stream write node knobs are always relevant, I'd move it higher up outside of these section.

BigRoy: Like this:

image

Note that I liked "Additional Knobs" better then ;)

And possibly move them higher up if they are always relevant:

image

jakubjezek001: the approach is solid, thanks @BigRoy

BigRoy: if it helps @moonyuet I can push this layout change in settings (since I prototyped it). Might save you a few minutes.

moonyuet: > if it helps @moonyuet I can push this layout change in settings (since I prototyped it). Might save you a few minutes.

Thanks!

BigRoy: https://github.com/ynput/ayon-nuke/pull/344/commits/aecf38f7d9fcacd5fc89a11e0d0a12c1fca2e31e

May need some cleanup - not sure if I refactored all that was needed. You'll need to still implement settings conversion to move the `extension` value. And of course change `client` code to use the new keys.

Also, please update PR description to make sure it matches latest code.

moonyuet: Settings overrides: [665641d](https://github.com/ynput/ayon-nuke/pull/344/commits/665641d0527ac3c56fc8d7261c3d51db7ddd0440)

use write_knobs as settings: [5c84ebb](https://github.com/ynput/ayon-nuke/pull/344/commits/5c84ebb2eceecc6d8398df9bf3dd87add28513b0)

client/ayon_nuke/api/plugin.py
moonyuet: [5c84ebb](https://github.com/ynput/ayon-nuke/pull/344/commits/5c84ebb2eceecc6d8398df9bf3dd87add28513b0)

client/ayon_nuke/api/plugin.py
moonyuet: [5c84ebb](https://github.com/ynput/ayon-nuke/pull/344/commits/5c84ebb2eceecc6d8398df9bf3dd87add28513b0)

client/ayon_nuke/api/plugin.py
moonyuet: [5c84ebb](https://github.com/ynput/ayon-nuke/pull/344/commits/5c84ebb2eceecc6d8398df9bf3dd87add28513b0)

server/settings/publish_plugins.py
moonyuet: [6e7cac4](https://github.com/ynput/ayon-nuke/pull/344/commits/6e7cac45973907f9df86d411a4fdeb94e4da640b)

client/ayon_nuke/api/plugin.py
```diff
@@ -1407,30 +1409,21 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
-            write_node["mov64_write_timecode"].setValue(1)
-        except Exception:
-            self.log.info("`mov64_write_timecode` knob was not found")
+        if self.custom_write_knobs:
+            for knob in self.custom_write_knobs:
+                if knob["name"] in {
+                    "file", "file_type",
+                    "channels", "raw", "mov64_fps"
+                }:
+                    continue
```
BigRoy: Should we disallow this already on the settings level as well?

And if we do skip something that is set in settings - I think we should really log a warning.

Also, why can't we set `mov64_fps` here? Why must we ignore it here to set it later?

I feel like we should just be doing it in this order.
```
self._set_defaults(write_node)
self._set_custom_knobs(write_node)
```
So that if some knob is file type specific then it's set after file type was set?

server/settings/conversion.py
```diff
@@ -297,6 +297,24 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+        .get("ExtractReviewIntermediates", {})
+        .get("outputs")
+    )
+    for output in extract_review_outputs:
+        extension = output.pop("extension", None)
+        if not extension:
+            # Nothing to convert
+            return
+        write_knobs = output.get("write_knobs")
+        write_knobs["file_type"] = extension
```
BigRoy: This doesn't work if `write_knobs` does not yet exist on the output.
This should be:

```
write_knobs = output.setdefault("write_knobs", dict())
write_knobs["file_type"] = extension
```

moonyuet: [181f8e6](https://github.com/ynput/ayon-nuke/pull/344/commits/181f8e62d17eca33413acd970a4a75adc6720da2)

client/ayon_nuke/api/plugin.py
moonyuet: [a0fc87d](https://github.com/ynput/ayon-nuke/pull/344/commits/a0fc87dcbda48b03f6a479ce046aad7e4accb805) and [1ec26f9](https://github.com/ynput/ayon-nuke/pull/344/commits/1ec26f9bb2dfafed508a31d3da37081b15508e17)

BigRoy: > Should we disallow this already on the settings level as well?
> 
> And if we do skip something that is set in settings - I think we should really log a warning.
> 
> Also, why can't we set mov64_fps here? Why must we ignore it here to set it later?

We're still skipping `mov64_fps` in the custom knobs logic and I'm a bit confused as to why, etc. because it's not really commented in the code. Why is that needed? 🤔

moonyuet: We skip this to follow the fps preset in the entities from the project.

server/settings/conversion.py
```diff
@@ -297,6 +297,25 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+        .get("outputs")
+    )
+    for output in extract_review_outputs:
+        extension = output.pop("extension", None)
+        if not extension:
+            # Nothing to convert
+            return
+
+        write_knobs = output.setdefault("write_knobs", {})
+        write_knobs["file_type"] = extension
```
jakubjezek001, line 320: this is not adding also the default set of `write_knobs` knobs for mov. It should really be adding also the prores related codec knobs

moonyuet: [a1b1956](https://github.com/ynput/ayon-nuke/pull/344/commits/a1b1956202f0a8935545f9ae7a5f9328d8cbb9d4)

moonyuet:
@jakubjezek001 @rdelillo @BigRoy can you review? Thanks

server/settings/conversion.py
```diff
@@ -297,6 +313,33 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+        if extension != "mov":
+            continue
+
+        custom_knobs = write_knobs.get("custom")
+        if custom_knobs:
+            continue
+
+        write_knobs["custom"] = _get_custom_knobs_list()
```
BigRoy: This is wrong. We should not be enforcing this here.

What @jakubjezek001 pointed out should really be settings conversion. With this logic here you won't be able to actually clear the additional knobs, which kind of beats the purpose of them being additional.

Also, we may want to change the key to `additional` instead of `custom` now that we're labeling them Additional Knobs in settings.

We need the settings conversion logic to be backwards compatible to this: https://github.com/ynput/ayon-nuke/pull/344/changes#diff-2b72fd20b424aa2837f801f8df5ae6fca06e7ea79c0d6b852c0cc72f65eb56daL1422-L1431 to ensure we remain setting the same things. But we should allow the admin to remove them if he/she wants.

BigRoy (changes requested):

server/settings/conversion.py
rdelillo: Maybe I miss something, but having them as default should be enough, no ?
We default to what used to be hardcoded, then admin can edit from there.

rdelillo (commented):
Tested locally and it worked with `mxf`, `mov` and `jpg`.
Note that I had an issue with single-frame image sequence in `Extract Review` but I don't think it comes from this PR more that it revealed it.

```
DEBUG: New representation tags: `['baking', 'review', 'ftrackreview', 'kitsureview', 'webreview']`
  File "C:\Users\robin\AppData\Local\Ynput\AYON\dependency_packages\ayon_2607151723_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 173, in process
    self.main_process(instance)
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 407, in main_process
    self._render_output_definitions(
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 482, in _render_output_definitions
    temp_data = self.prepare_temp_data(instance, repre, output_def)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 710, in prepare_temp_data
    input_frames = list(sorted(cols[0].indexes))
                               ~~~~^^^
IndexError: list index out of range
```

[publish-report-260810-15-08.json](https://github.com/user-attachments/files/30910527/publish-report-260810-15-08.json)

Once Roy's feedback is addressed, I'm happy to approve and merge this one if this needs to be tackled as a separate PR.

server/settings/conversion.py
moonyuet: I agree with @rdelillo, we can add the data as default settings and allow admin to edit it, and we won't need to store the extra piece of data for just the conversion. It is actually already at the default settings. https://github.com/ynput/ayon-nuke/blob/a1b1956202f0a8935545f9ae7a5f9328d8cbb9d4/server/settings/publish_plugins.py#L489-L508

BigRoy: > Maybe I miss something, but having them as default should be enough, no ?

It is not.

Here's an example, in older settings I had overrides to Extract Review Intermediates, something as simple as just the Output name of the baking stream changed from `baking` to `bakinga`.

Then copy settings to this new addon. I get this:

image

Which means on copying from older settings if I had overrides, I'm losing the Additional Knobs defaults. We should be applying them also on the copy settings in that case. So that the publish behaves the same as it did before.

As such, we must apply these defaults also on copy settings to enforce these into the `custom` key accordingly.

@moonyuet

server/settings/conversion.py
```diff
@@ -297,6 +297,39 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+    for output in extract_review_outputs:
+        extension = output.pop("extension", None)
+        if not extension:
+            # Nothing to convert
+            return
+
+        write_knobs = output.setdefault("write_knobs", {})
+        write_knobs["file_type"] = extension
+
+        if "custom" not in write_knobs or write_knobs["custom"] is None:
```
BigRoy:
```suggestion
        if write_knobs.get("custom") is None:
```

server/settings/conversion.py
```diff
@@ -297,6 +297,39 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
     ] = sync_workfile_version_on_product_base_types
 
 
+def _convert_review_intermediates_model_0_4_11(overrides: dict) -> None:
+    """Convert review intermediates extension model to include
```
BigRoy: For our sanity in the future, should we do a check at the start here that the source version was `<= 0.4.11`?

Just to ensure below only triggers if we copy from older settings.

E.g. like:

- https://github.com/ynput/ayon-core/blob/2adee9a0f020ba928912d91fa8f08ab5a9d64915/server/settings/conversion.py#L300-L301
- https://github.com/ynput/ayon-houdini/blob/e12da7d7d8e40737184b1864dbb3e1950e8c71f9/server/settings/conversion.py#L41-L42

server/settings/conversion.py
moonyuet: @BigRoy @jakubjezek001 let me know if it is still wrong in terms of overrides. [6bbc40e](https://github.com/ynput/ayon-nuke/pull/344/commits/6bbc40e2f7a20cae41d3b292ad96a0e841e7c085)

server/settings/conversion.py
```diff
@@ -297,6 +297,39 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+                    "type": "text",
+                    "name": "mov64_codec",
+                    "text": "ap4h",
+                },
+                {
+                    "type": "boolean",
+                    "name": "mov64_write_timecode",
+                    "boolean": True,
+                },
+            ]
```
BigRoy, line 334: So if an admin adds a new "output baking stream" - do we want that added entry to also default to having this set? if so, then we must supply these default values also via the default factory or default on the `custom` settings field.

I think we should, right @jakubjezek001 ?

It'd be similar to this: https://github.com/ynput/ayon-usd/blob/e2b6cdd01df2326bc0b062447e136b91f5eccab0/server/settings/main.py#L218-L223 where you'd pass a list of the settings fields definition as 'default' value that represent the matching settings.

Which I think would be like:
```python
    custom: list[KnobModel] = SettingsField(
        title="Additional Knobs",
        default=[
            KnobModel(
                type="text",
                name="mov64_codec",
                text="ap4h",
            ),
            KnobModel(
                type="boolean",
                name="mov64_write_timecode",
                boolean=True,
            )
        ]
    )
```
But I can't seem to get this to work.

I noticed now that we're also not doing these "defaults" for the Reformat Nodes > Reposition Knobs for added baking stream entries.

BigRoy (commented):
Settings conversion now works, nice! :)

I do have some remaining mostly cosmetic notes.

server/settings/conversion.py
jakubjezek001: yes, please, if it is possible it would be good to have it as default. User can change those values if needed

server/settings/conversion.py
```diff
@@ -297,10 +298,47 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
     ] = sync_workfile_version_on_product_base_types
 
 
+def _convert_review_intermediates_model_0_4_11(
+        overrides: dict, version: VersionInfo) -> None:
+    """Convert review intermediates extension model to include
+    file_type extension."""
+    if (version.major, version.minor, version.patch) >= (0, 4, 11):
```
BigRoy: 0.4.11 is already released, and we'll want to support converting from that version.
```suggestion
    if (version.major, version.minor, version.patch) > (0, 4, 11):
```

server/settings/conversion.py
BigRoy: @moonyuet see if you can make this work, otherwise check with Martin/Luke how settings should be defined to make this possible.

server/settings/conversion.py
moonyuet: [e95926b](https://github.com/ynput/ayon-nuke/pull/344/commits/e95926be863f8a3271e2056b166ecc50e4117673)

server/settings/conversion.py
moonyuet: [e95926b](https://github.com/ynput/ayon-nuke/pull/344/commits/e95926be863f8a3271e2056b166ecc50e4117673)

moonyuet: > @moonyuet see if you can make this work, otherwise check with Martin/Luke how settings should be defined to make this possible.

It is working for default setting but not the custom added stream setting..

server/settings/conversion.py
```diff
@@ -297,10 +300,51 @@ def _convert_collect_sync_workfile_version_model_0_4_5(overrides: dict) -> None:
...
+                        name="mov64_codec",
+                        text="ap4h",
+                    ),
+                    KnobModel(
+                        type="boolean",
+                        name="mov64_write_timecode",
+                        boolean=True,
+                    )
+                ]
+            )
```
BigRoy: Revert this conversion to use the dictionaries please @moonyuet. It'll keep the code maintenance easier. Why?

Over time `custom` attribute's model may change (e.g. the `KnobModel` might take different arguments) - but the settings conversion from older versions should still do the same thing. If we are now referencing the live settings model, then suddenly upon a future change of the settings model we'll also need to adapt the EARLIER settings conversion which at that point will become really hard to test over time. If we use plain dicts for that, we don't have that problem.

server/settings/publish_plugins.py
```diff
@@ -146,6 +146,19 @@ class ReformatNodesConfigModel(BaseSettingsModel):
     )
 
 
+class WriteKnobsModel(BaseSettingsModel):
+    file_type: str = SettingsField(False)
+    custom: list[KnobModel] = SettingsField(
+        default_factory=list,
+        title="Additional Knobs"
+    )
```
BigRoy, line 165: This @moonyuet is where we'd want to be able to specify the default. So that if someone adds a new baking stream that contains the `WriteKnobsModel` that it defaults to ALSO having the defaults of this `WriteKnobsModel` with the list prefilled. Along the lines of:
```
    custom: list[KnobModel] = SettingsField(
        title="Additional Knobs",
        default=[
            KnobModel(
                type="text",
                name="mov64_codec",
                text="ap4h",
            ),
            KnobModel(
                type="boolean",
                name="mov64_write_timecode",
                boolean=True,
            )
        ]
    )
```
[src](https://github.com/ynput/ayon-nuke/pull/344#discussion_r3757107162)


But like you said that doesn't seem to work on the frontend. The question is, are we using it wrong - or is just not working or supported on the backend/frontend?

server/settings/conversion.py
BigRoy: Let's move this discussion here: https://github.com/ynput/ayon-nuke/pull/344#discussion_r3758358228 because it's closer to where that change would need to go.

server/settings/publish_plugins.py
moonyuet: Sorry, I misunderstood a little bit for this. I thought you are talking about conversion.py but actually you are talking about the default custom knobs settings in `publish_plugins.py`. But if there is a way to expose the setting in ayon_usd, we can also expose it in here.

server/settings/conversion.py
moonyuet: [7f0b654](https://github.com/ynput/ayon-nuke/pull/344/commits/7f0b6543b2705be471e06f920586e8551d74601e)

server/settings/publish_plugins.py
moonyuet: This commit [5e422ea](https://github.com/ynput/ayon-nuke/pull/344/commits/5e422eac9d84776f5842d9fe6ee104a1730c0884) should be carrying the default settings, but you need to save your stream name to activate it.

BigRoy: In that case - this hints at a frontend bug @Innders

That's definitely what I wanted, but then without it requiring to first save for frontend to actually show these values :D

BigRoy: So the settings here are indeed correct - it seems to be a backend bug because with these changes in backend: https://github.com/ynput/ayon-backend/pull/1059 then suddenly it does live show them.

client/ayon_nuke/api/plugin.py
```diff
@@ -1485,6 +1466,46 @@ def _connect_to_above_nodes(self, node, product_name, message):
...
+        for knob in custom_knobs:
+            if knob["name"] in {
+                "file", "file_type",
+                "channels", "raw", "mov64_fps"
+            }:
+                log.warning(
+                    f"Skipping to set custom knob '{knob['name']}' "
+                    "as it is already set in defaults."
+                )
+                continue
```
rdelillo: I think there are different treatments here:
* `file`, `file_type` which is already set from settings in default
* `channel`, `mov64_fps` which is also set in default but enforced from instance data
* `raw` always enfored as True from default

```suggestion
            if knob["name"] in {
                "file", "file_type",
            }:
                log.debug(
                    f"Skipping '{knob['name']}' as it is already set."
                )
                continue
           
            elif knob["name"] in {
                "channels", "raw", "mov64_fps"
            }:
                log.warning(
                    f"Ignoring provided custom knob '{knob['name']}' "
                    "as it is already enforced from instance data."
                )
                continue
```

Now on the workflow side here, it feels a bit weird to me that we provide custom write knobs settings but ignore those who don't please our defaults. Meaning you cannot create 2 review intermediates at different fps.

From a dev/pipeline TD, I'd rather expect we set our default, then update the write knobs from the settings, then whatever happens next is user responsability.

client/ayon_nuke/api/plugin.py
```diff
@@ -1485,6 +1466,46 @@ def _connect_to_above_nodes(self, node, product_name, message):
...
+            to_type = knob["type"]
+            value = convert_knob_value_to_correct_type(
+                to_type, knob[to_type]
+            )
+            write_node[knob["name"]].setValue(value)
```
rdelillo: Do we check that `knob["name"]` exists on `write_node` first ? 
If it comes from settings there might be typo.

client/ayon_nuke/api/plugin.py
moonyuet: >From a dev/pipeline TD, I'd rather expect we set our default, then update the write knobs from the settings, then whatever happens next is user responsability.

You have the point though. Maybe allowing TDs to customize their setup so that it would be easier to do testing. I will remove (despite I just commit your suggestion just now). Removed in [174f0b3](https://github.com/ynput/ayon-nuke/pull/344/commits/174f0b3d50f7b719f3fd362f070938bedafb2cfa)

client/ayon_nuke/api/plugin.py
moonyuet: [174f0b3](https://github.com/ynput/ayon-nuke/pull/344/commits/174f0b3d50f7b719f3fd362f070938bedafb2cfa)

rdelillo (approved):
Works as expected, tested with multiple review intermediate and did not fail on non-existing knobs 🥳 .

Image

There is the default on new settings and single-frame review intermediate bugs but they should be handled via other PRs.

BigRoy (approved):

PR #347, opened 2026-08-05
Respect display and view in ExporterReview

jakubjezek001:
## Changelog Description
ExporterReview now preserves OCIO display and view data separately when baking reviews in Nuke. This enables downstream integration to derive the correct colorspace template value for display/view transforms instead of losing the selected view metadata.

## Additional info
Representation colorspace data now accepts `display` and `view`. The dependent `ayon-core` change derives the integration template value from these fields when no direct colorspace is available.

## Testing notes:
At your Intermediate baking preset use dispaly and view from your active OCIO config. Ideally test at ACES 1.3.
The outcome should look as expected.

## Dependency
Depends on https://github.com/ynput/ayon-core/pull/1983

### DIFF
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1473,7 +1473,7 @@ class ExporterReviewMov(ExporterReview):
         if delete:
             tags.append("delete")
 
-        if config_data and not display:
+        if config_data and (not colorspace and not display):
             # backward compatibility: convert display and view to colorspace
             # just in case older nuke_default ocio config
             colorspace = get_display_view_colorspace_name(
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1473,7 +1473,7 @@ class ExporterReviewMov(ExporterReview):
         if delete:
             tags.append("delete")
 
-        if not display:
+        if config_data and not display:
             # backward compatibility: convert display and view to colorspace
             # just in case older nuke_default ocio config
             colorspace = get_display_view_colorspace_name(
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -1020,6 +1020,7 @@ class ExporterReview(object):
                 "display": display,
                 "view": view,
             })
+
         self.data["representations"].append(repre)
 
     def get_imageio_baking_profile(self):
@@ -1471,6 +1472,15 @@ class ExporterReviewMov(ExporterReview):
 
         if delete:
             tags.append("delete")
+
+        if not display:
+            # backward compatibility: convert display and view to colorspace
+            # just in case older nuke_default ocio config
+            colorspace = get_display_view_colorspace_name(
+                config_path=config_data["path"],
+                display=display,
+                view=view
+            )
 
         self.get_representation_data(
             tags=tags + add_tags,
--- a/client/ayon_nuke/api/plugin.py
+++ b/client/ayon_nuke/api/plugin.py
@@ -947,18 +947,24 @@ class ExporterReview(object):
         range=False,
         custom_tags=None,
         colorspace=None,
+        display=None,
+        view=None,
     ):
-        """Add representation data to self.data
+        """Add representation data to self.data.
 
         Args:
             tags (list[str], optional): list of defined tags.
-                                        Defaults to None.
+                Defaults to None.
             range (bool, optional): flag for adding ranges.
-                                    Defaults to False.
+                Defaults to False.
             custom_tags (list[str], optional): user inputted custom tags.
-                                               Defaults to None.
+                Defaults to None.
             colorspace (str, optional): colorspace name.
-                                        Defaults to None.
+                Defaults to None.
+            display (str, optional): display name.
+                Defaults to None.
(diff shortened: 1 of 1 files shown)


client/ayon_nuke/api/plugin.py
```diff
@@ -1001,6 +1007,20 @@ def get_representation_data(
...
+                self.instance.context.data,
+                colorspace=display,
+                log=self.log
+            )
+            # remove colorspace key from colorspaceData
+            repre["colorspaceData"].pop("colorspace", None)
+            repre["colorspaceData"].update({
+                "display": display,
+                "view": view,
+            })
```
moonyuet, line 1022: Would it be more explicit to adjust the colorspace data from the instance.data to collect display and view instead of doing in representation data?

BigRoy: The most accurate would be if we always store it in representation display view I think, _if we can_. Because that way different representations can each have a different colorspace or display/view set instead of all representations sharing the one.

BigRoy (commented):
We'll need to be wary of a few things @jakubjezek001 

It looks like we're taking the display/view from the Viewer LUT and not from the write node. What makes it so in Nuke that the write node actually includes that display/view in the written out image? Or is that display/view NOT written into the pixel data of the image that Nuke writes out?

Because there's a few things we should consider:

- The output image may or may not have a display/view baked in, the **source display**
- The user did see in their scene a display/view that the output image may not get baked in but still wants to apply to reviewables, the **scene display**
- We may want to convert it to some display/view, the **target display** (although this is usually more up to e.g. Extract Review or alike to decide.)

See the terminology here: https://github.com/ynput/ayon-core/pull/1541#issue-3611957112

The question here then is, are we referring here to the **Scene Display** (what the user was seeing in Nuke through the viewer) or are we referring to the **Source Display** (what the image written out by Nuke got baked in.)

I'd say whenever we put data on `representation` as display/view it must really represent the source display/view. Putting scene display/view on the representation makes little scene I'd argue. We may also be referring to the "target display/view" but I haven't seen a case yet where we'd want that explicitly defined at this level PER representation perhaps?


@MustafaJafar just so you're aware ☝️ - we may want to make sure this terminology is clear in our docs somehow. And that perhaps certain settings like those for Extract OIIO Transcode or alike could perhaps use as well in settings so it's clear that you're setting either target display/view, etc.

jakubjezek001:
> It looks like we're taking the display/view from the Viewer LUT and not from the write node. What makes it so in Nuke that the write node actually includes that display/view in the written out image? Or is that display/view NOT written into the pixel data of the image that Nuke writes out?

Yes you are not completely wrong. The display/view is actually baked in pixels, that is why it is added to representation's colorspaceData so down the line it can be used as sourceDisplay/View. Here we are applying for backward compatibility with older nuke versions - where write node did not support display/view knobs, applying it via OCIODisplay node above the write node and setting to RAW colorspace just to disable any colorspace baking by the write node.

client/ayon_nuke/api/plugin.py
BigRoy: @jakubjezek001 can you make a follow up issue to core to allows also setting `display` and `view` via `set_colorspace_data_to_representation`?

client/ayon_nuke/api/plugin.py
```diff
@@ -1456,11 +1473,22 @@ def generate_mov(self, farm=False, delete=True, **kwargs):
...
             tags.append("delete")
 
+        if config_data and (not colorspace and not display):
+            # backward compatibility: convert display and view to colorspace
+            # just in case older nuke_default ocio config
+            colorspace = get_display_view_colorspace_name(
+                config_path=config_data["path"],
+                display=display,
+                view=view
+            )
```
BigRoy, line 1483: So display would be empty or none here? does that mean that only `view` is set? Because it requires colorspace to not be set and display to not be set?

BigRoy (commented):
Not a crazy approach ;) so please validate it does what you'd expect in testing, and then fine with me.

client/ayon_nuke/api/plugin.py
jakubjezek001: yes nuke_default ocio config is making **display** only optional and **view** will always be set. For this case the **colorspace** will be resolve from the **view** only input, since it will be resolved form the mentioned ocio config. This will only provide the condition in the case, other should always be including values in both **display/view** or with **colorspace** directly

jakubjezek001:
> Not a crazy approach ;) so please validate it does what you'd expect in testing, and then fine with me.

thank you, but in the slate at ACES 1.3 this did not fix the issue. It seems the Display colorspace in ACES 1.3 are really correctly working as RRT and ODT and this was already provided by the code before the fixes. My openion is that this enhancement is still good to be merged since it makes more sense and is more explicit. But without it, nothing is broken.

BigRoy:
> > Not a crazy approach ;) so please validate it does what you'd expect in testing, and then fine with me.
> 
> thank you, but in the slate at ACES 1.3 this did not fix the issue. It seems the Display colorspace in ACES 1.3 are really correctly working as RRT and ODT and this was already provided by the code before the fixes. My openion is that this enhancement is still good to be merged since it makes more sense and is more explicit. But without it, nothing is broken.

Let's go through that next week then please.

PR #353, opened 2026-08-19
Add `check_inventory_version` in create and load callback function

### LINKED ISSUE #352: `check_inventory_versions` should be called on workfile load (via workfiles app)

## Current Behavior
check_inventory_versions is currently only being called when opening a fresh nuke instance after a workfile is loaded - or when manually opening a script via file-open.
When using the workfiles app from an open nuke session the function is not being called after the workfile is loaded.

## Expected Behavior
It should be called when using the workfiles app to open a workfile.

### Steps To Reproduce:

1. Open nuke 
2. Open a workfile using the workfiles app.
3. check_inventory_versions is not being called.

### Version
Ayon-Nuke 0.4.12

Windows

### Additional information
I debugged a little bit and it seems that these lines do not have any effect:

https://github.com/ynput/ayon-nuke/blob/6eda0511776c2b7b789f76237f2be491fd868bfc/client/ayon_nuke/api/pipeline.py#L145-L146

Furthermore, this
https://github.com/ynput/ayon-nuke/blob/6eda0511776c2b7b789f76237f2be491fd868bfc/client/ayon_nuke/api/pipeline.py#L247
function is being called after a workfile is opened via the workfiles app.

Not sure what happens under the hood and why "on_script_load" is not being called,
but we could just implement **check_inventory_versions** in both "on_script_load" and "on_root_create" and optionally create a server setting for it.
(And get rid of the broken "workio.open_file" event callback?)

moonyuet:
## Changelog Description
This PR is to add `check_inventory_version` in create and load callback function to ensure the check_inventory_version would apply when the nuke session is opened via workfile tools.

Fix https://github.com/ynput/ayon-nuke/issues/352#

## Testing notes:
1. Launch Nuke 
2. Load something with not the latest version
3. Save your scene via workfile tools(SceneA)
4. Updated to the latest version to your loaded asset
5. Save your scene via workfile tools (SceneB)
6. Open your Scene A back via workfile tools
7. It should be showing the correct color.

### DIFF
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -242,29 +242,33 @@ def add_nuke_callbacks(project_settings: dict = None):
 
 
 def on_root_create() -> None:
-    """Callback function for on script create."""
-    # set apply all workfile settings on script load and save
-    workfile_settings = WorkfileSettings()
-    on_script_create_settings = (
-        workfile_settings.project_settings
-        ["nuke"]
-        ["workfile_callbacks"]
-        ["on_script_create"]
-    )
-
-    if on_script_create_settings["set_resolution"]:
-        workfile_settings.reset_resolution()
-
-    if on_script_create_settings["set_frame_range"]:
-        workfile_settings.reset_frame_range_handles()
-
-    if on_script_create_settings["set_colorspace"]:
-        workfile_settings.set_colorspace()
-
+    """Callback function for on root node create."""
     # adding favorites to file browser
+    workfile_settings = WorkfileSettings()
     workfile_settings.set_favorites()
-    # template builder callbacks
-    start_workfile_template_builder()
+
+    # Root created callback is also triggered on scene load because that also
+    # creates a new root node. So we check if current file is None to run
+    # logic that we only want to apply on new scene creation.
+    if current_file() is None:
+        on_script_create_settings = (
+            workfile_settings.project_settings
+            ["nuke"]
+            ["workfile_callbacks"]
+            ["on_script_create"]
+        )
+
+        if on_script_create_settings["set_resolution"]:
+            workfile_settings.reset_resolution()
+
+        if on_script_create_settings["set_frame_range"]:
+            workfile_settings.reset_frame_range_handles()
+
+        if on_script_create_settings["set_colorspace"]:
+            workfile_settings.set_colorspace()
+
+        # template builder callbacks
+        start_workfile_template_builder()
 
 
 def on_script_load() -> None:
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -1,7 +1,6 @@
 import nuke
 
 import os
-import contextlib
 import importlib
 from collections import OrderedDict, defaultdict
 
@@ -18,7 +17,7 @@ from ayon_core.host import (
 
 from ayon_core.host.interfaces import SaveWorkfileContext
 
-from ayon_core.lib import register_event_callback, Logger
+from ayon_core.lib import Logger
 from ayon_core.pipeline import (
     register_loader_plugin_path,
     register_creator_plugin_path,
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -226,11 +226,8 @@ def add_nuke_callbacks(project_settings: dict = None):
 
     nuke_settings = project_settings["nuke"]
 
-    # Set all workfile settings.'
     nuke.addOnCreate(on_root_create, nodeClass="Root")
-    # set checker for last versions on loaded containers
-    nuke.addOnScriptLoad(check_inventory_versions)
-    # fix ffmpeg settings on script
+
     nuke.addOnScriptLoad(on_script_load)
 
     # set checker for last versions on loaded containers
@@ -268,6 +265,8 @@ def on_root_create() -> None:
     workfile_settings.set_favorites()
     # template builder callbacks
     start_workfile_template_builder()
+    # set checker for last versions on loaded containers
+    check_inventory_versions()
 
 
 def on_script_load() -> None:
@@ -297,6 +296,9 @@ def on_script_load() -> None:
 
     if on_script_load_settings["set_colorspace"]:
         workfile_settings.set_colorspace()
+
+    # set checker for last versions on loaded containers
+    check_inventory_versions()
 
 
 def reload_config():
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -239,7 +239,6 @@ def add_nuke_callbacks(project_settings: dict = None):
 
 def on_root_create() -> None:
     """Callback function for on root node create."""
-    log.info("On root create...")
     # adding favorites to file browser
     workfile_settings = WorkfileSettings()
     workfile_settings.set_favorites()
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -164,7 +164,6 @@ class NukeHost(
         root_node = nuke.root()
         set_node_data(root_node, ROOT_DATA_KNOB, data)
 
-
     def _before_workfile_save(
         self, save_workfile_context: SaveWorkfileContext
     ) -> None:
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -265,8 +265,6 @@ def on_root_create() -> None:
     workfile_settings.set_favorites()
     # template builder callbacks
     start_workfile_template_builder()
-    # set checker for last versions on loaded containers
-    check_inventory_versions()
 
 
 def on_script_load() -> None:
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -1,6 +1,7 @@
 import nuke
 
 import os
+import contextlib
 import importlib
 from collections import OrderedDict, defaultdict
 
@@ -142,9 +143,6 @@ class NukeHost(
         register_inventory_action_path(INVENTORY_PATH)
         register_workfile_build_plugin_path(WORKFILE_BUILD_PATH)
 
-        # Register AYON event for workfiles loading.
-        register_event_callback("workio.open_file", check_inventory_versions)
-
     def setup_ui_callbacks_and_menu(self):
         """Setup AYON menus."""
         if not nuke.GUI:
@@ -243,6 +241,7 @@ def add_nuke_callbacks(project_settings: dict = None):
 
 def on_root_create() -> None:
     """Callback function for on root node create."""
+    log.info("On root create...")
     # adding favorites to file browser
     workfile_settings = WorkfileSettings()
     workfile_settings.set_favorites()
@@ -251,6 +250,7 @@ def on_root_create() -> None:
     # creates a new root node. So we check if current file is None to run
     # logic that we only want to apply on new scene creation.
     if current_file() is None:
+        log.info("On new file..")
         on_script_create_settings = (
             workfile_settings.project_settings
             ["nuke"]
@@ -273,6 +273,7 @@ def on_root_create() -> None:
 
 def on_script_load() -> None:
     """Callback function for on script load."""
+    log.info("On script load...")
     # fix ffmpeg settings on script
     if nuke.env["LINUX"]:
         nuke.tcl('load ffmpegReader')
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -247,7 +247,6 @@ def on_root_create() -> None:
     # creates a new root node. So we check if current file is None to run
     # logic that we only want to apply on new scene creation.
     if current_file() is None:
-        log.info("On new file..")
         on_script_create_settings = (
             workfile_settings.project_settings
             ["nuke"]
--- a/client/ayon_nuke/api/workio.py
+++ b/client/ayon_nuke/api/workio.py
@@ -30,6 +30,10 @@ def open_file(filepath):
             nuke.Root()["name"].setValue(nuke_script)
             nuke.Root()["project_directory"].setValue(os.path.dirname(nuke_script))
             nuke.Root().setModified(False)
+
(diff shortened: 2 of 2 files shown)


BigRoy (changes requested):
I understand the reasoning here, but I think it's the wrong fix. I wouldn't expect `on_script_create` to get called when opening a file.

I think the reason why the nuke load event does not trigger and hence `on_script_load` does not trigger is because of how we open the file here:

https://github.com/ynput/ayon-nuke/blob/8a401a7360ecd3c99143e175af7e047630074ec5/client/ayon_nuke/api/workio.py#L29

Essentially, we're not opening a new script - but we're reading it - which is a workaround we have to avoid Nuke opening a new window. Which makes me think we should enforce, from that open logic to trigger the load callback manually then.

So that whenever we do a load, we still expect the load callback to be triggerered.

Looking at that logic it only applies `if not ASSIST` so it may need to be at the end of that `if` with some comment as to why we need to trigger the event ourselves.

I'd actually almost go as far as to just put a call to [`nuke.onScriptLoad()`](https://learn.foundry.com/nuke/developers/140/pythonreference/_autosummary/nuke.onScriptLoad.html#nuke-onscriptload) there.


So, confirm this is indeed the case, and if so - add the dedicated call to `nuke.onScriptLoad()` to enforce it.

BigRoy:
@moonyuet I noticed you already re-requested review, but this wasn't added yet - right?

> I'd actually almost go as far as to just put a call to [nuke.onScriptLoad()](https://learn.foundry.com/nuke/developers/140/pythonreference/_autosummary/nuke.onScriptLoad.html#nuke-onscriptload) [there](https://github.com/ynput/ayon-nuke/blob/8a401a7360ecd3c99143e175af7e047630074ec5/client/ayon_nuke/api/workio.py#L29).

moonyuet:
> @moonyuet I noticed you already re-requested review, but this wasn't added yet - right?
> 
> > I'd actually almost go as far as to just put a call to [nuke.onScriptLoad()](https://learn.foundry.com/nuke/developers/140/pythonreference/_autosummary/nuke.onScriptLoad.html#nuke-onscriptload) [there](https://github.com/ynput/ayon-nuke/blob/8a401a7360ecd3c99143e175af7e047630074ec5/client/ayon_nuke/api/workio.py#L29).

so we dont need this PR basically, right?

If you talk about adding the script load callback in the open workfile, we might need to find if there is some alternatives for that like the callback for workfile

sebastianbrandhuber:
@moonyuet @BigRoy 
Just saw the discussion here.
What about we manually trigger all existing **onScriptLoad** callbacks right after here?

https://github.com/ynput/ayon-nuke/blob/8a401a7360ecd3c99143e175af7e047630074ec5/client/ayon_nuke/api/workio.py#L29

Something like 
```
callbacks = [tup[0] for cb_list in nuke.onScriptLoads.values() for tup in cb_list]
for func in callbacks:
    if callable(func):
        func()
```
should do the trick.

The problem is, that **nuke.addOnCreate** callbacks will be executed aswell and most likely execute some functions twice.
We might need to clear **nuke.onCreates** before executing the code and restore it afterwards, but I did not test that part.

BigRoy:
> What about we manually trigger all existing **onScriptLoad** callbacks right after here?

No need, you can just call `nuke.onScriptLoad()` ;) 

> The problem is, that nuke.addOnCreate callbacks will be executed aswell and most likely execute some functions twice.
> We might need to clear nuke.onCreates before executing the code and restore it afterwards, but I did not test that part.

I think we shouldn't actually. I think we should make our "open file" behavior mimic as if the user would do File > Open manually as much as possible. it seems like, if the user does that - it also triggers both the on create root callback and the script open callback.

The assumption that on create would only happen on "new" files and not on scene open then maybe is wrong. 

I've now pushed two commits that should make it behave as we'd expect.

client/ayon_nuke/api/workio.py
```diff
@@ -30,6 +30,10 @@ def read_script(nuke_script):
             nuke.Root()["name"].setValue(nuke_script)
             nuke.Root()["project_directory"].setValue(os.path.dirname(nuke_script))
             nuke.Root().setModified(False)
+
+            # Above way of loading script does not trigger onScriptLoad()
+            # callback, so we need to call it manually
+            nuke.onScriptLoad()
```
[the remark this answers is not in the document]
BigRoy: > instead of relying on an undocumented/ambiguous Nuke API entry point.

It's in their API docs? 🤔 And we intentionally use their call because we want callbacks that some TD or plug-in may have assigned to nuke to also trigger, because this should really behave as a script open to the users.

client/ayon_nuke/api/pipeline.py
```diff
@@ -245,29 +242,33 @@ def add_nuke_callbacks(project_settings: dict = None):
...
-        ["on_script_create"]
-    )
+    workfile_settings.set_favorites()
 
-    if on_script_create_settings["set_resolution"]:
-        workfile_settings.reset_resolution()
+    # Root created callback is also triggered on scene load because that also
+    # creates a new root node. So we check if current file is None to run
+    # logic that we only want to apply on new scene creation.
+    if current_file() is None:
```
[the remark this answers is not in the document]
BigRoy: That's correct - and indeed not the behavior we want. We actually _don't_ want "new file" logic to get triggered when we open a file. best would be if we can actually disable the callbacks on that first clear call because we don't want anything responding to that as if it's temporarily creating an empty scene 🤔 

So that's fixed with: https://github.com/ynput/ayon-nuke/pull/353/commits/79b242640f27521680c95ee3c2d571449cc21a1f

BigRoy:
I pushed some tweaks @moonyuet.

Essentially now it should behave similar to regular file open as to the callbacks that get called.
- It doesn't trigger the onCreate callback anymore before file open, only after it's been opened.
- It also triggers the onScriptLoad callback now.
- I've also changed the autosave prompt logic - previously it first opened the file, then asked to open the autosave, and then opened the autosave. Essentially meaning it'd open the file twice if you clicked yes - now it'll give the prompt before opening the file, and hence open only once.
- I believe this now also doesn't trigger the empty scene on create logic before the file is opened when opening a file on launch, so it doesn't run logic potentially twice that would've been discarded due to the scene file being opened anyway.

Things to test would be:
- Workfile template builder still runs when it should be running.
- callbacks trigger on the right moments.
- callbacks should not be running multiple times unnecessarily if nuke would usually not do that either on e.g. file open (like on startup or autosave as described above)

moonyuet:
> I pushed some tweaks @moonyuet.
> 
> Essentially now it should behave similar to regular file open as to the callbacks that get called.
> 
> * It doesn't trigger the onCreate callback anymore before file open, only after it's been opened.
> * It also triggers the onScriptLoad callback now.
> * I've also changed the autosave prompt logic - previously it first opened the file, then asked to open the autosave, and then opened the autosave. Essentially meaning it'd open the file twice if you clicked yes - now it'll give the prompt before opening the file, and hence open only once.
> * I believe this now also doesn't trigger the empty scene on create logic before the file is opened when opening a file on launch, so it doesn't run logic potentially twice that would've been discarded due to the scene file being opened anyway.
> 
> Things to test would be:
> 
> * Workfile template builder still runs when it should be running.
> * callbacks trigger on the right moments.
> * callbacks should not be running multiple times unnecessarily if nuke would usually not do that either on e.g. file open (like on startup or autosave as described above)

Thanks first of all. You commit lots of changes.

I am actually thinking much simpler approach to do. We need to investigate why this: https://github.com/ynput/ayon-nuke/blob/6eda0511776c2b7b789f76237f2be491fd868bfc/client/ayon_nuke/api/pipeline.py#L145-L146 this line of code is not working. And make it work and then it should be working.
It should be
```python
register_event_callback("workfile.open", check_inventory_versions)
```
This is the video for the code changes and it should be working to display the container color right, unless I missed something which is not making the inventory check working.
https://github.com/user-attachments/assets/3e8881da-d82c-4a96-bd01-5ed46be01504

BigRoy:
> I am actually thinking much simpler approach to do. We need to investigate why this:
> 
> https://github.com/ynput/ayon-nuke/blob/6eda0511776c2b7b789f76237f2be491fd868bfc/client/ayon_nuke/api/pipeline.py#L145-L146
> 
> this line of code is not working. And make it work and then it should be working.

The event should have been `workfile.open.after`: https://github.com/ynput/ayon-core/blob/688610322a00c34557ff6893de9919cc3cc174b4/client/ayon_core/host/interfaces/workfiles.py#L1799-L1810

However, the issue with that is that we're still triggering the onCreate callback on an empty scene, so we have a brief moment we're triggering logic against an empty scene... then we're loading the file, but not actually doing the OTHER events of the on script load. (which is fixable by connecting `on_script_load` to that event of course) but the bigger issue I had with that approach is that we're fixing our AYON callbacks only, but our open file still does not feel like it's close to a native open file method. So if some studio has existing logic that triggers on nuke open file callbacks with the on script load events of Nuke itself then they suddenly don't trigger via the AYON workfiles. I think that's a massive oversight. Because a studio may just rely on a plug-in or pre-built script that does something with these events in a useful manner; and we don't want to be telling all of them that they need to adapt their scripts just because we're not doing a regular file open.

In your case, you're solely fixing that the inventory logic triggers which is what the issue describes is a resulting problem, but it's not the root issue we're dealing with. It's that our logic is not emitting the right events to begin with.

If you'd like, I recommend adding a `log.info` call to each of our registered events, and see when they actually run. And you'll see that e.g. on create was also triggering on file open via workfiles tool (and hence triggering the workfile template builder!) because we do the script clear. This may massively affect performance if that's the case.

We need to confirm:
- We don't trigger redundant logic on file open, e.g. it should not initialize a new scene in-between like e.g. workfile template builder should not trigger or setting context options on the empty file should not be done if we're doing a file open, because we do that after the read file.
- The callbacks behave closely like File > Open (behavior is same with native Nuke logic)
- Custom callbacks a studio may have to `nuke.addOnScriptLoad` should preferably also trigger with our workfile tool opens, because to the user/admin - that's just a regular file open.

moonyuet:
@BigRoy I tested the code in my side. It seems working fine. But as mentioned in the call, we need to deal with the issue when`on_root_create` performed when we close the script/close the application. We can leave it as a separate issue.

@rdelillo @jakubjezek001 can you take a look at this code? Thanks

BigRoy:
> ut as mentioned in the call, we need to deal with the issue when`on_root_create` performed when we close the script/close the application. We can leave it as a separate issue.

Yup, just to explain this issue.

Whenever in GUI mode and you close nuke, then before it closes it actually clears the current comp and that triggers a root node creation and hence that callback. That in turn does some of our set resolution logic, etc. but also trigger workfile template build if that's configured. Something that would be really good to avoid if we can.

client/ayon_nuke/api/workio.py
```diff
@@ -21,40 +35,69 @@ def save_file(filepath):
...
+            root["project_directory"].setValue(os.path.dirname(nuke_script))
+            root.setModified(False)
+
+            # Above way of loading script does not trigger onCreate() nor
+            # onScriptLoad() callback, so we need to call it manually here
+            # after root name has been set so the callbacks behave similar
+            # to how they would if user would do File > Open or on Nuke
+            # launch with a comp script file.
+            nuke.onCreate()
+            nuke.onScriptLoad()
```
rdelillo, line 79: Testing this PR made me realize we have an inconsistency beween Nuke and Nuke Assist here.

Nuke and Nuke-X opens in place but Nuke Assist opens in a separate window:
This was done to ensure Nuke Assist could load script from Nuke instead of failing https://github.com/ynput/ayon-nuke/commit/4f4d589c828a105abb143bf1cb18f2caf945d386
@jakubjezek001 is that fine ?

@moonyuet could we use this PR to better document Nuke Assist exception please ?
Current comment line 98 makes it feel like it will always open in place when it does not.

rdelillo (approved):
Approving this one as it works as expected in Nuke and Nuke Assist.
(I could also replicate the `on_root_create` bug mentioned above).

client/ayon_nuke/api/workio.py
moonyuet: I elaborated the comment in [dd9eab3](https://github.com/ynput/ayon-nuke/pull/353/commits/dd9eab359eb81d7f8a25472f7ebe81fd3cfffc3d)

BigRoy: The comment added in https://github.com/ynput/ayon-nuke/pull/353/commits/dd9eab359eb81d7f8a25472f7ebe81fd3cfffc3d doesn't really tell me anything. It explains what `read_script` does which if we're explaining that should be the docstring of that function instead. But the point that Robin makes here, is that there's difference in how Nuke Assist does it... which this comment now still does not explain. The comment should really be about _why_ it differs for Nuke Assist. What is making Nuke Assist special that we can't use the same behavior. Also, that comment should also be inside `read_script` before the if branch.

moonyuet: [58efb15](https://github.com/ynput/ayon-nuke/pull/353/commits/58efb153cb85c7235f93db897fc0c46b9b1dd2fd) I moved the comment to read_script. Let me know it still doesn't explain. Maybe docstring would help too.

BigRoy: Yes, this is better 👍

jakubjezek001:
@BigRoy please once you approve you can merge it, thank you

client/ayon_nuke/api/pipeline.py
```diff
@@ -245,33 +238,40 @@ def add_nuke_callbacks(project_settings: dict = None):
 
 
 def on_root_create() -> None:
-    """Callback function for on script create."""
-    # set apply all workfile settings on script load and save
+    """Callback function for on root node create."""
+    log.info("On root create...")
```
BigRoy:
```suggestion
```

client/ayon_nuke/api/pipeline.py
```diff
@@ -245,33 +238,40 @@ def add_nuke_callbacks(project_settings: dict = None):
...
-    )
+    workfile_settings.set_favorites()
 
-    if on_script_create_settings["set_resolution"]:
-        workfile_settings.reset_resolution()
+    # Root created callback is also triggered on scene load because that also
+    # creates a new root node. So we check if current file is None to run
+    # logic that we only want to apply on new scene creation.
+    if current_file() is None:
+        log.info("On new file..")
```
BigRoy:
```suggestion
```

BigRoy (approved):
OK for me - may just require the logs to be removed?

PR #358, opened 2026-08-20
Remove backwards compatibility with Nuke 13 and below

BigRoy:
## Changelog Description

Support only Nuke 14+ (older versions are using Python lower than Py3.9 anyway)

Drop support for older versions. Not required to merge soon because having the code doesn't break anything. Just keeping this around until we're absolutely certain we've dropped the older apps anyway.

- Foundry Nuke 13.2 uses Python 3.7.7
- Foundry Nuke 14.0 uses Python 3.9.1

## Testing notes:

1. Nuke 14+ should still work as intended.


jakubjezek001 (approved):
tested on Nuke 17 and works well

PR #360, opened 2026-08-21
[YN-1042] Templated Builder: options to trigger when application launch or always on new scene

### LINKED ISSUE #359: YN-1042 - New Workfile Template Builder System for Nuke
Regarding to the implementation of the new workfile template builder in Core, Nuke workfile template builder needs refactoring to synchronize with the new settings(in spite of the old function still acts as backward compatibility for use)

_No response_

### Additional context:
[YN-1042](https://support.ayon.app/projects/Active_Clients/overview/projects/Active_Clients/overview/projects/Active_Clients?project=Active_Clients&type=folder&id=4693fb93e95549098ea8f6248533ea9f)

moonyuet:
## Changelog Description
This PR is to implement the new templated builder system in Nuke to support trigger when application launch or always on new scene.
Resolve https://github.com/ynput/ayon-nuke/issues/359

Related to https://github.com/ynput/ayon-core/pull/1907

## Testing notes:
1. Tweak the settings in ayon+settings://blender/templated_workfile_build/profiles/0/
2. Build workfile

### DIFF
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -62,6 +62,7 @@ from .constants import (
 )
 
 from .utils import get_node_outputs
+from .workio import current_file
 
 from .colorspace import get_formatted_display_and_view
 
@@ -2734,16 +2735,16 @@ def prompt_reset_context():
 
 
 def start_workfile_template_builder() -> None:
-    """trigger workfile template builder for every nuke new file if enabled."""
-    # remove callback since it would be duplicating the workfile
-    nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")
-    # Ignore opened scripts: only run new-file trigger for unsaved scenes.
-    if nuke.root().name().lower() != "root":
+    """Trigger workfile template builder when a new file is created."""
+    if current_file() is not None:
         return
 
-    log.info("Starting workfile template builder...")
-    from .workfile_template_builder import trigger_on_app_launch
-    trigger_on_app_launch()
+    from .workfile_template_builder import trigger_on_new_file
+    host = registered_host()
+    log.info(f"App initialized: {host.app_initialized}")
+    if not host.app_initialized:
+        log.info("Triggering workfile template builder on new file...")
+        trigger_on_new_file()
 
 
 def add_scripts_menu():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2492,9 +2492,13 @@ def maintained_selection(exclude_nodes=None):
         # unselect all selection in case there is some
         reset_selection()
 
-        # and select all previously selected nodes
-        if previous_selection:
-            select_nodes(previous_selection)
+        for node in previous_selection:
+            try:
+                node["selected"].setValue(True)
+            except ValueError:
+                # Could fail if a node was deleted during the context.
+                # See: https://github.com/ynput/ayon-nuke/pull/331 for details.
+                pass
 
 
 @contextlib.contextmanager
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2747,10 +2747,9 @@ def start_workfile_template_builder() -> None:
         log.info("Triggering workfile template builder on application launch.")
         trigger_on_app_launch()
         os.environ["AYON_NUKE_INITIALIZED"] = "True"
-        return
-
-    log.info("Triggering workfile template builder on new file.")
-    trigger_on_new_file()
+    else:
+        log.info("Triggering workfile template builder on new file.")
+        trigger_on_new_file()
 
 
 def add_scripts_menu():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -15,9 +15,7 @@ from qtpy import QtCore, QtWidgets
 import ayon_api
 
 from ayon_core.host import HostDirmap
-from ayon_core.pipeline.workfile.workfile_template_builder import (
-    TemplateProfileNotFound
-)
+
 from ayon_core.lib import (
     env_value_to_bool,
     Logger,
@@ -2735,20 +2733,14 @@ def prompt_reset_context():
         dialog.deleteLater()
 
 
-def start_workfile_template_builder():
-    from .workfile_template_builder import (
-        build_workfile_template
-    )
+def start_workfile_template_builder() -> None:
+    """trigger workfile template builder for every nuke launch if enabled."""
+    from .workfile_template_builder import trigger_on_app_launch
+    # Ignore opened scripts: only run new-file trigger for unsaved scenes.
+    if nuke.root().name().lower() != "root":
+        return
 
-    # remove callback since it would be duplicating the workfile
-    nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")
-
-    # to avoid looping of the callback, remove it!
-    log.info("Starting workfile template builder...")
-    try:
-        build_workfile_template(workfile_creation_enabled=True)
-    except TemplateProfileNotFound:
-        log.warning("Template profile not found. Skipping...")
+    trigger_on_app_launch()
 
 
 def add_scripts_menu():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -62,7 +62,6 @@ from .constants import (
 )
 
 from .utils import get_node_outputs
-from .workio import current_file
 
 from .colorspace import get_formatted_display_and_view
 
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2734,24 +2734,6 @@ def prompt_reset_context():
         dialog.deleteLater()
 
 
-def start_workfile_template_builder() -> None:
-    """Trigger workfile template builder when a new file is created."""
-    if current_file() is not None:
-        return
-
-    from .workfile_template_builder import (
-        trigger_on_app_launch,
-        trigger_on_new_file,
-    )
-    if not os.getenv("AYON_NUKE_INITIALIZED"):
-        log.info("Triggering workfile template builder on application launch.")
-        trigger_on_app_launch()
-        os.environ["AYON_NUKE_INITIALIZED"] = "True"
-    else:
-        log.info("Triggering workfile template builder on new file.")
-        trigger_on_new_file()
-
-
 def add_scripts_menu():
     try:
         from scriptsmenu import launchfornuke
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2747,7 +2747,6 @@ def start_workfile_template_builder() -> None:
         log.info("Triggering workfile template builder on application launch.")
         trigger_on_app_launch()
         os.environ["AYON_NUKE_INITIALIZED"] = "True"
-        return
 
     log.info("Triggering workfile template builder on new file.")
     trigger_on_new_file()
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2736,6 +2736,8 @@ def prompt_reset_context():
 
 def start_workfile_template_builder() -> None:
     """trigger workfile template builder for every nuke new file if enabled."""
+    # remove callback since it would be duplicating the workfile
+    nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")
     # Ignore opened scripts: only run new-file trigger for unsaved scenes.
     if nuke.root().name().lower() != "root":
         return
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -78,6 +78,7 @@ EXCLUDED_KNOB_TYPE_ON_READ = (
 JSON_PREFIX = "JSON:::"
 ROOT_DATA_KNOB = "publish_context"
 INSTANCE_DATA_KNOB = "publish_instance"
+app_launch = True
 
 
 class DeprecatedWarning(DeprecationWarning):
@@ -2735,12 +2736,21 @@ def prompt_reset_context():
 
 def start_workfile_template_builder() -> None:
     """trigger workfile template builder for every nuke new file if enabled."""
-    from .workfile_template_builder import trigger_on_new_file
     # Ignore opened scripts: only run new-file trigger for unsaved scenes.
     if nuke.root().name().lower() != "root":
         return
+    global app_launch
+    from .workfile_template_builder import (
+        trigger_on_app_launch,
+        trigger_on_new_file
+    )
+    if app_launch:
+        trigger_on_app_launch()
+        app_launch = False
+        return
 
-    trigger_on_new_file()
+    if not app_launch:
+        trigger_on_new_file()
 
 
 def add_scripts_menu():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -2734,13 +2734,13 @@ def prompt_reset_context():
 
 
 def start_workfile_template_builder() -> None:
-    """trigger workfile template builder for every nuke launch if enabled."""
-    from .workfile_template_builder import trigger_on_app_launch
+    """trigger workfile template builder for every nuke new file if enabled."""
+    from .workfile_template_builder import trigger_on_new_file
     # Ignore opened scripts: only run new-file trigger for unsaved scenes.
     if nuke.root().name().lower() != "root":
         return
 
-    trigger_on_app_launch()
+    trigger_on_new_file()
 
 
 def add_scripts_menu():
--- a/client/ayon_nuke/api/lib.py
+++ b/client/ayon_nuke/api/lib.py
@@ -78,7 +78,6 @@ EXCLUDED_KNOB_TYPE_ON_READ = (
 JSON_PREFIX = "JSON:::"
 ROOT_DATA_KNOB = "publish_context"
 INSTANCE_DATA_KNOB = "publish_instance"
-app_launch = True
 
 
 class DeprecatedWarning(DeprecationWarning):
@@ -2741,18 +2740,10 @@ def start_workfile_template_builder() -> None:
     # Ignore opened scripts: only run new-file trigger for unsaved scenes.
     if nuke.root().name().lower() != "root":
         return
-    global app_launch
-    from .workfile_template_builder import (
-        trigger_on_app_launch,
-        trigger_on_new_file
-    )
-    if app_launch:
-        trigger_on_app_launch()
-        app_launch = False
-        return
 
-    if not app_launch:
-        trigger_on_new_file()
+    log.info("Starting workfile template builder...")
+    from .workfile_template_builder import trigger_on_app_launch
+    trigger_on_app_launch()
 
 
 def add_scripts_menu():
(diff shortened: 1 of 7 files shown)


BigRoy:
This may need `package.py` requirements to core version as well.

moonyuet:
> This may need `package.py` requirements to core version as well.

Indeed. but later. I didn't fully implement the stuff yet and would like to have discussion for that.

BigRoy:
This PR is probably best made on top of https://github.com/ynput/ayon-nuke/pull/353 because that changes quite a bit what triggers where on launch/new files.

moonyuet:
@rdelillo @jakubjezek001 @BigRoy 
https://github.com/ynput/ayon-nuke/blob/1437a4cc2bf50fe0c8954f8d883ec629fe0b6a1c/client/ayon_nuke/api/lib.py#L2738-L2751

I am thinking if we should just instead of using getter function. We should just simply make use of this condition to decide the action would be triggered as based on the comment. But this also means, we need to move the code to the `pipeline.py`
Some pseudo code.
```python

def start_workfile_template_builder():
	 global app_launch
    from .workfile_template_builder import (
        NukeTemplateBuilder
    )
	builder = NukeTemplateBuilder(registered_host())
	preset = builder.get_template_preset()
    # remove callback since it would be duplicating the workfile
    nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")

    # to avoid looping of the callback, remove it!
	if preset.execute_on_new_file:
		log.info("Starting workfile template builder...")
    	builder.build_template()
	if  execute_on_app_launch and app_launch and not preset.execute_on_new_file:
		log.info("Starting workfile template builder...")
		builder.build_template()
		app_launch = False

```

moonyuet:
I committed [48567f5](https://github.com/ynput/ayon-nuke/pull/360/commits/48567f5d9e364cba82422f44ab7481e53e9f62f2) after several testing. 
I haven't implemented the `trigger_on_new_file` as the workfile template system appears to act a bit differently in terms of nuke. Nuke has their own logics in terms of the way it deals with new scene.(Launch new application process for new scene)
If we launched nuke, it launches the application process.
If we click `File -> New Comp` in Nuke, it will launch a new application process.
Trigger on application launch means we would build the builder every time nuke launched. 
@BigRoy @jakubjezek001 @rdelillo any thoughts on that?

LiborBatek (changes requested):
Im getting load failures when trying to open the template file....

I have double checked it and it really points to the nuke `.nk` file 

Image

Image

server/settings/templated_workfile_build.py
```diff
@@ -21,9 +21,13 @@ class TemplatedWorkfileProfileModel(BaseSettingsModel):
     keep_placeholder: bool = SettingsField(
         False,
         title="Keep placeholders")
+    execute_on_app_launch: bool = SettingsField(
```
BigRoy, line 29: Are we lacking a `execute_on_new_file`?

moonyuet: See https://github.com/ynput/ayon-nuke/pull/360#issuecomment-5391966641

LiborBatek:
@moonyuet it does not error out while trying to open the template file but nothing happens...neither building nor opening the actual template file works.

I guess I should wait for new commits still, or? I have also tried different template file to work with but not helped neither.

moonyuet:
> @moonyuet it does not error out while trying to open the template file but nothing happens...neither building nor opening the actual template file works.
> 
> 
> 
> I guess I should wait for new commits still, or? I have also tried different template file to work with but not helped neither.

Does your workile template exist? I have added a check on to see if it is valid path, and if the path doesn't exist it won't do anything.

BigRoy:
> If we click `File -> New Comp` in Nuke, it will launch a new application process.
> Trigger on application launch means we would build the builder every time nuke launched.

It'll open a new Nuke window, but I'd say that from the perspective of the workfile template builder, we should look at that as a "new file" if we can and not a "new app launch". So that pressing CTRL+N can still be defined to not enforce the template build. The question is whether we can support that from Nuke's integration perspective. I suppose we can?

LiborBatek:
@moonyuet now after the latest commits gives me again the load error while trying to open the template file...

image

LiborBatek (approved):
Finally with `ayon-core` recent [PR 2018](https://github.com/ynput/ayon-core/pull/2018) merged it works...

I was able to use the `Build from template` action and also `Open Template` action successfully!

There is only one caveat, that `Apply on New Scene` does not have effect and template will be used always no matter what. I guess its current Nuke limitation?!

Other than that it works well!

BigRoy:
> @moonyuet now after the latest commits gives me again the load error while trying to open the template file...
> 
> image

The issue here is that there's a space in front of `C:\...` right? Or at least this error message makes it look like so.

client/ayon_nuke/api/lib.py
```diff
@@ -80,6 +78,7 @@
 JSON_PREFIX = "JSON:::"
 ROOT_DATA_KNOB = "publish_context"
 INSTANCE_DATA_KNOB = "publish_instance"
+app_launch = True
```
BigRoy: Might need a comment to explain why we're not storing it on the registered host? I assume it's because Nuke may re-initialize AYON when pressing CTRL+N (which admittedly is a thing we may want to solve at some point as well.)

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,25 @@ def prompt_reset_context():
...
 
-def start_workfile_template_builder():
-    from .workfile_template_builder import (
-        build_workfile_template
-    )
-
+def start_workfile_template_builder() -> None:
+    """trigger workfile template builder for every nuke new file if enabled."""
     # remove callback since it would be duplicating the workfile
     nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")
```
BigRoy: Pretty sure this does nothing now - because we're not adding this callback anywhere anymore?

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,25 @@ def prompt_reset_context():
...
+    if nuke.root().name().lower() != "root":
+        return
+    global app_launch
+    from .workfile_template_builder import (
+        trigger_on_app_launch,
+        trigger_on_new_file
+    )
+    if app_launch:
+        trigger_on_app_launch()
+        app_launch = False
```
BigRoy: Is it really on to `start_workfile_template_builder` to set `app_launch` variable to False, should we not do that on the first run inside host install itself or alike which actually handles the app launch?

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,25 @@ def prompt_reset_context():
...
+        return
 
-    # to avoid looping of the callback, remove it!
-    log.info("Starting workfile template builder...")
-    try:
-        build_workfile_template(workfile_creation_enabled=True)
-    except TemplateProfileNotFound:
-        log.warning("Template profile not found. Skipping...")
+    if not app_launch:
```
BigRoy: This can be `else:`

client/ayon_nuke/api/workfile_template_builder.py
```diff
@@ -160,9 +161,28 @@ def collect_placeholders(self):
...
+    """
+    builder = NukeTemplateBuilder(registered_host())
+    builder.trigger_on_new_file()
+
+
+def build_workfile_template():
+    builder = NukeTemplateBuilder(registered_host())
+    preset = builder.get_template_preset()
+    if not preset.has_valid_path():
+        return
```
BigRoy: When running `build_workfile_template` from the AYON menu I'd expect an invalid path to be reported as error?

BigRoy (commented):
> [There is only one caveat, that Apply on New Scene does not have effect and template will be used always no matter what. I guess its current Nuke limitation?!](https://github.com/ynput/ayon-nuke/pull/360#pullrequestreview-5031281904)

This may need some extra checking @moonyuet because that should work fine, right?

moonyuet:
> > @moonyuet now after the latest commits gives me again the load error while trying to open the template file...
> > image
> 
> The issue here is that there's a space in front of `C:\...` right? Or at least this error message makes it look like so.

I had a call with Libor yesterday. It is not about whether there is a space in the front of `C:\` but it is about `C:\Project\template.nk` where the `\t` from `\template.nk` can be confusing as trailing tab character in text.

moonyuet:
> > [There is only one caveat, that Apply on New Scene does not have effect and template will be used always no matter what. I guess its current Nuke limitation?!](https://github.com/ynput/ayon-nuke/pull/360#pullrequestreview-5031281904)
> 
> This may need some extra checking @moonyuet because that should work fine, right?

I will remove the apply on new scene then or use just apply on new scene for nuke workfile template. We should make sure the workfile template works in respect to how nuke integrates with AYON. Will be happy if anyone can come up with better plan on this. 
We do `always on new scene` where when users click CTRL + N or File -> Comp for new comp or launch nuke application they build workfile template.  or We use `Launch on Nuke App` for that name. 
Thoughts? @rdelillo @jakubjezek001 @BigRoy

client/ayon_nuke/api/workfile_template_builder.py
moonyuet: [992d5c3](https://github.com/ynput/ayon-nuke/pull/360/commits/992d5c307c2b0bac95bb2792f44d2151cc6ed16f)

client/ayon_nuke/api/lib.py
jakubjezek001: Arent we adding it here? `ayon_nuke/api/pipeline.py`

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,25 @@ def prompt_reset_context():
...
-def start_workfile_template_builder():
-    from .workfile_template_builder import (
-        build_workfile_template
-    )
-
+def start_workfile_template_builder() -> None:
+    """trigger workfile template builder for every nuke new file if enabled."""
     # remove callback since it would be duplicating the workfile
     nuke.removeOnCreate(start_workfile_template_builder, nodeClass="Root")
+    # Ignore opened scripts: only run new-file trigger for unsaved scenes.
```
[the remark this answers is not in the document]
moonyuet: @jakubjezek001 should we add `start_workfile_template_builder` respectively from other onCreate callback?

jakubjezek001:
> I will remove the apply on new scene then or use just apply on new scene for nuke workfile template. We should make sure the workfile template works in respect to how nuke integrates with AYON. Will be happy if anyone can come up with better plan on this. We do `always on new scene` where when users click CTRL + N or File -> Comp for new comp or launch nuke application they build workfile template. or We use `Launch on Nuke App` for that name. Thoughts? @rdelillo @jakubjezek001 @BigRoy

Here is mine suggestion:
- run template only on App launch
- create new workfile from the template only if settings toggle is enabled `Save first workfile version`
- do not run workfile template if CTRL-N is used for new window
- do not run tempate creation if a version already exists

moonyuet:
> > I will remove the apply on new scene then or use just apply on new scene for nuke workfile template. We should make sure the workfile template works in respect to how nuke integrates with AYON. Will be happy if anyone can come up with better plan on this. We do `always on new scene` where when users click CTRL + N or File -> Comp for new comp or launch nuke application they build workfile template. or We use `Launch on Nuke App` for that name. Thoughts? @rdelillo @jakubjezek001 @BigRoy
> 
> Here is mine suggestion:
> 
> * run template only on App launch
> * create new workfile from the template only if settings toggle is enabled `Save first workfile version`
> * do not run workfile template if CTRL-N is used for new window
> * do not run tempate creation if a version already exists

I pushed some commits as showed above.

client/ayon_nuke/api/lib.py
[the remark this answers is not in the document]
moonyuet: [e39e17e](https://github.com/ynput/ayon-nuke/pull/360/commits/e39e17e86e902ae48f02f2889169af1277370c9e)

BigRoy:
> * do not run tempate creation if a version already exists

Unfortunately, this is **NOT** how workfile template builder behaves in general. We shouldn't define a custom behavior for Nuke. All DCCs must behave the same.

The user configures: "does this run on launch?" If enabled, then it does if you're opening into an empty scene (not opening a file on launch). Then usually with the new settings in core we also have a "does this run on new file?" - if we can't support that inside Nuke, then I guess we remove that one... but pretty sure we can detect it fine, because we can detect the new windows, which in Nuke's view is just a new scene.

> create new workfile from the template only if settings toggle is enabled Save first workfile version

Again, that's a **separate** setting. We shouldn't implement Nuke-specific behavior. I think it's worth for @MustafaJafar to write documentation for all these settings so that we have a go to resource of the intended behavior. But the way I see it:

- trigger on app launch: whether template should build when starting app without a workfile
- trigger on new file: whether "new scene" inside the DCC initializes with a template build
- save first version: on app launch, whether to save the first workfile version (only applies if there's no existing version yet.)

Or at least, that's my understanding of how it's working across the integrations?

In short, make sure it behaves like all the other integrations ;)

jakubjezek001:
> The user configures: "does this run on launch?" If enabled, then it does if you're opening into an empty scene (not opening a file on launch). Then usually with the new settings in core we also have a "does this run on new file?" - if we can't support that inside Nuke, then I guess we remove that one... but pretty sure we can detect it fine, because we can detect the new windows, which in Nuke's view is just a new scene.

In that case we are missing nuke profile toggle `execute_on_new_file`. And we can actually provide it, since nuke is labeling every new scene/script which is not saved with explicit name. Those are titled as `Untitled` and api is returning `Root` on them. There is actually function in `client/ayon_nuke/api/workio.py` called **current_file** and if it returns **None** then it is fresh new scene/script.

client/ayon_nuke/api/pipeline.py
```diff
@@ -265,9 +265,7 @@ def on_root_create() -> None:
         if on_script_create_settings["set_colorspace"]:
             workfile_settings.set_colorspace()
 
-        # template builder callbacks
-        start_workfile_template_builder()
-
+    start_workfile_template_builder()
```
BigRoy: This should still be in the indent, right? Because we only ever care about calling this if we're not in an active workfile?

client/ayon_nuke/api/pipeline.py
```diff
@@ -784,3 +779,21 @@ def select_instance(instance):
...
+        print(">>> [ App initialized: True ]")
+        print(">>> [ Triggering workfile template builder on application launch. ]")
+        # Your first-time initialization code here
+        _app_initialized = True
+        return True
+    else:
+        # Subsequent initializations (New Comp)
+        print(">>> [ App initialized: False ]")
+        # Your subsequent initialization code here (if any)
+        return False
```
BigRoy: Can be removed? Or at least, appears unused?

client/ayon_nuke/api/workfile_template_builder.py
```diff
@@ -4,11 +4,19 @@
...
     PlaceholderPlugin,
     PlaceholderItem
 )
 
+try:
+    from ayon_core.tools.workfile_template_build import open_template_ui
+
+except ImportError:
+    def open_template_ui(*_args, **_kwargs):
+        raise RuntimeError("Template UI is not available in non-GUI mode.")
```
BigRoy: What's this? Why do we need to safeguard this? Should we safeguard this in `open_template` call below instead?

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,22 @@ def prompt_reset_context():
...
-        build_workfile_template(workfile_creation_enabled=True)
-    except TemplateProfileNotFound:
-        log.warning("Template profile not found. Skipping...")
+    if not os.getenv("AYON_NUKE_INITIALIZED"):
+        log.info("Triggering workfile template builder on application launch.")
+        trigger_on_app_launch()
+        os.environ["AYON_NUKE_INITIALIZED"] = "True"
+    else:
+        log.info("Triggering workfile template builder on new file.")
+        trigger_on_new_file()
```
BigRoy: I'd argue to just remove this function - and just put this logic that is here directly around line 267 in `pipeline.py` inside the if not current file indentation?

Also, if we're triggering the workfile template build - does it even still make sense to trigger the reset of workfile settings before it? Aren't we then just resetting things, to then just open another scene file? Or? 🤔

client/ayon_nuke/api/lib.py
```diff
@@ -2735,20 +2734,22 @@ def prompt_reset_context():
...
-    # to avoid looping of the callback, remove it!
-    log.info("Starting workfile template builder...")
-    try:
-        build_workfile_template(workfile_creation_enabled=True)
-    except TemplateProfileNotFound:
-        log.warning("Template profile not found. Skipping...")
+    if not os.getenv("AYON_NUKE_INITIALIZED"):
+        log.info("Triggering workfile template builder on application launch.")
+        trigger_on_app_launch()
+        os.environ["AYON_NUKE_INITIALIZED"] = "True"
```
BigRoy: Would it be feasible to handle the setting of this inside the host install method or whatever we consider is the actual initialization of the AYON Nuke? It being in the workfile template build logic feels off.

BigRoy (changes requested):

client/ayon_nuke/api/lib.py
moonyuet: > Would it be feasible to handle the setting of this inside the host install method or whatever we consider is the actual initialization of the AYON Nuke? It being in the workfile template build logic feels off.

I have tried to implement the variable in the host (as if other hosts to check whether the application is initialized). it doesn't work as it should be. Maybe there is better way to fix it but currently I can't think of any. I have added comments to explain why the environment variable is there. 
[8487668](https://github.com/ynput/ayon-nuke/pull/360/commits/84876681525ff2ab8e1b97af35c42f1ab59a012f)

client/ayon_nuke/api/lib.py
moonyuet: > Also, if we're triggering the workfile template build - does it even still make sense to trigger the reset of workfile settings before it? Aren't we then just resetting things, to then just open another scene file? Or? 🤔

Application launch: open scene files only application launch

Always new scene: open scene files only when always new scene(File-> New Comp and Ctrl + N would trigger the workfile build)

So it should be working as if other hosts does already

client/ayon_nuke/api/workfile_template_builder.py
moonyuet: [597c611](https://github.com/ynput/ayon-nuke/pull/360/commits/597c6111e5fe87dbcb8d4a61d2cb5406f1ca9993)

client/ayon_nuke/api/lib.py
moonyuet: [85f0e6f](https://github.com/ynput/ayon-nuke/pull/360/commits/85f0e6fd91c46f7225cad1645a7bf85d48fa6028)

client/ayon_nuke/api/pipeline.py
moonyuet: [85f0e6f](https://github.com/ynput/ayon-nuke/pull/360/commits/85f0e6fd91c46f7225cad1645a7bf85d48fa6028)

client/ayon_nuke/api/pipeline.py
moonyuet: [85f0e6f](https://github.com/ynput/ayon-nuke/pull/360/commits/85f0e6fd91c46f7225cad1645a7bf85d48fa6028)

jakubjezek001:
changing this to review ready makes sense, since we are already reviewing it for some time ;)

LiborBatek (approved):
I have tested multiple scenarios and all of these were working as expected...

LGTM

Been tested with `Nuke 16`

BigRoy (approved):

Issue #357, opened 2026-08-20
Root node create is triggered when closing the application


moonyuet:
> we need to deal with the issue when`on_root_create` performed when we close the script/close the application. We can leave it as a separate issue.

Whenever in GUI mode and you close nuke, then before it closes it actually clears the current comp and that triggers a root node creation and hence that callback. That in turn does some of our set resolution logic, etc. but also trigger workfile template build if that's configured. Something that would be really good to avoid if we can.

_Originally posted by @BigRoy in https://github.com/ynput/ayon-nuke/issues/353#issuecomment-5355096784_

## Current Behavior

As came up [here](https://github.com/ynput/ayon-nuke/pull/353#issuecomment-5355096784) whenever you shut down Nuke (when running in GUI mode) when you close down the app before it shuts down it clears the current comp and apparently generates a new root node, which our `nuke.addOnCreate` registered callback acts upon.

## Expected Behavior

When Nuke is being shut down, we should not be triggering any logic that we intended for new files.

moonyuet:
The script below can fix the issue:
```python
def add_nuke_callbacks(project_settings: dict = None):
    """Adding all available nuke callbacks"""
    if project_settings is None:
        project_settings = get_current_project_settings()

    nuke_settings = project_settings["nuke"]

    # Set all workfile settings.'
    nuke.addOnCreate(on_root_create, nodeClass="Root")
    # set checker for last versions on loaded containers
    nuke.addOnScriptLoad(check_inventory_versions)
    # fix ffmpeg settings on script
    nuke.addOnScriptLoad(on_script_load)

    # set checker for last versions on loaded containers
    nuke.addOnScriptSave(check_inventory_versions)

    nuke.addOnScriptClose(deregister_nuke_callbacks)

    if nuke_settings["dirmap"]["enabled"]:
        log.info("Added Nuke's dir-mapping callback ...")
        # Add dirmap for file paths.
        nuke.addFilenameFilter(dirmap_file_name_filter)

    log.info("Added Nuke callbacks ...")


def deregister_nuke_callbacks():
    nuke.removeOnCreate(on_root_create, nodeClass="Root")
```

BigRoy:
[...]

PR #362, opened 2026-08-25
add addOnScriptClose to deregistered on create nuke callbacks

moonyuet:
## Changelog Description
This PR is to avoid create callback won't be triggered after script closed.
Fixes https://github.com/ynput/ayon-nuke/issues/357

n/a

## Testing notes:
1. Launch Script A from Nuke
2. File -> New Comp and load Script B
3. Close Script A
4. Script B -> New Comp
5. All the callbacks still registered
6. Nuke app closed won't register the create callback anymore.

### DIFF
--- a/client/ayon_nuke/api/pipeline.py
+++ b/client/ayon_nuke/api/pipeline.py
@@ -222,12 +222,17 @@ def add_nuke_callbacks(project_settings: dict = None):
 
     nuke_settings = project_settings["nuke"]
 
+    # Set all workfile settings.'
     nuke.addOnCreate(on_root_create, nodeClass="Root")
-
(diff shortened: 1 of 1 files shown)


BigRoy (changes requested):
If you use File > Close Comp or CTRL+W shortcut the callbacks are now deregistered but you're actively in an empty comp. As such, unfortunately this may cause cases where you're working whilst no callbacks are registered.

Without this PR in the target branch it correctly invokes the new file / template builder in that scenario. In this PR it does not because the callbacks were deregistered.

moonyuet:
I closed it first as this blocks the other PR merged and we continue to discuss how to make it work.

PR #363, opened 2026-08-31
[YN-0692] Write node button: Single frame padding for image product

### LINKED ISSUE #247: YN-0692: Single frame in nuke read from rendered button missing padding issue

Write Render node button for Reading rendered output as read node is broken for Image still products. It is reporting that there is padding issue for rendered images. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0692]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=212462e0424711f1991d271f89000e98)

moonyuet:
## Changelog Description
This PR is to fix the single frame padding issue for image product when users use write node button.
Resolve https://github.com/ynput/ayon-nuke/issues/247

n/a

## Testing notes:
For each image, pre-render and render instance creators:
1. Create an single frame instance to get a write container
2. Render locally through the "Render local" button
3. Load back the result through "Read from Rendered"

### DIFF
--- a/client/ayon_nuke/startup/write_to_read.py
+++ b/client/ayon_nuke/startup/write_to_read.py
@@ -74,6 +74,7 @@ def detect_file_on_disk(
     collections, _ = clique.assemble(
         [combined_relative_path],
         assume_padded_when_ambiguous=True,
+        minimum_items=1,
         patterns=[clique.PATTERNS['frames']]
     )
 
@@ -88,26 +89,20 @@ def detect_file_on_disk(
 
         if files_in_dir:
             # Assemble all files in directory to find the complete sequence
-            coll, _ = clique.assemble(
+            coll = clique.assemble(
                 files_in_dir,
                 assume_padded_when_ambiguous=True,
                 patterns=[clique.PATTERNS['frames']]
             )[0][0]
             if coll.padding == collection.padding:
                 # Found matching sequence
-                directory_path = os.path.dirname(k_value)
-                basename = coll.head
-                extension = coll.tail
                 firstframe = min(coll.indexes)
                 lastframe = max(coll.indexes)
-                filepath = os.path.join(
-                    directory_path,
-                    f"{basename}{'#' * coll.padding}{extension}"
-                ).replace('\\', '/')
+                filepath = f"{coll.head}{'#' * coll.padding}{coll.tail}"
     # Convert to relative path if requested
     if filepath and allow_relative and project_dir:
         filepath = os.path.relpath(filepath, project_dir)
-        filepath = filepath.replace('\\', '/')
+    filepath = filepath.replace('\\', '/')
 
     return filepath, firstframe, lastframe
 
--- a/client/ayon_nuke/startup/write_to_read.py
+++ b/client/ayon_nuke/startup/write_to_read.py
@@ -14,7 +14,7 @@ def detect_file_on_disk(
         project_dir: str,
         first_frame: int,
         allow_relative: bool
-) -> tuple[str, int, int] | None:
+) -> tuple[str | None, int, int]:
     """Detects a file or image sequence on disk and returns its path along
     with the first and last frame numbers.
 
--- a/client/ayon_nuke/startup/write_to_read.py
+++ b/client/ayon_nuke/startup/write_to_read.py
@@ -2,6 +2,7 @@ import re
 import os
 import glob
 import nuke
+import clique
 from ayon_core.lib import Logger
 log = Logger.get_logger(__name__)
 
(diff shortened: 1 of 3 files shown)


client/ayon_nuke/startup/write_to_read.py
```diff
@@ -119,15 +119,33 @@ def write_to_read(gn,
...
-
             if n.knob('file') is not None:
-                myfile, firstFrame, lastFrame = evaluate_filepath_new(
-                    n.knob('file').getValue(),
-                    n.knob('file').evaluate(),
-                    project_dir,
-                    comp_start,
-                    allow_relative
-                )
+                if is_single_frame_output(n):
```
rdelillo, line 123: I don't think this is the fix.

Instead of bypassing `evaluate_filepath_new` with custom logic, I think we should be fix it so it handles single-frame sequence, still image and video files seemlessly. 

In my case the problem with still image came from:
```python
# IndexError when no frame number in path
frame_number = re.findall(r'\d+', combined_relative_path)[-1] 
```
But I can also see other pieces in current logic that look very fragile to say the least.

We should rewrite `evaluate_filepath_new` do be something like this:
```python
def detect_file_on_disk(...) -> Tuple[str, int, int] | None:
    #1- Check if expected filepath is frame-padded sequence (one or multiple files) or single file (image or video file)
    #2- Check if destination directory exists 
    #3a- Return path to single file if it exists, return None otherwise
    #3b- Check if sequence exists on disk with available frames, return padded str + available frame range or None otherwise
```

rdelillo (changes requested):
@moonyuet could you please elaborate a bit of the original issue ? Having the original traceback would help here.
E.g. from my tests it did not raise with frame padding problem but an `IndexError` on still image.

Also I think the testing notes need to be adjusted.
From your fix the problem isn't in publishing/rendering but in reading the result back. 
E.g. one should read:
```
For each image, pre-render and render instance creators:
1- Create an single frame instance to get a write container
2- Render locally through the "Render local" button
3- Load back the result through "Read from Rendered"
```

moonyuet:
> @moonyuet could you please elaborate a bit of the original issue ? Having the original traceback would help here. E.g. from my tests it did not raise with frame padding problem but an `IndexError` on still image.
> 
> Also I think the testing notes need to be adjusted. From your fix the problem isn't in publishing/rendering but in reading the result back. E.g. one should read:
> 
> ```
> For each image, pre-render and render instance creators:
> 1- Create an single frame instance to get a write container
> 2- Render locally through the "Render local" button
> 3- Load back the result through "Read from Rendered"
> ```

The original issue does not enclose any traceback and we need to test it by our own. I faced exactly the same exception error as you have faced. And I also face an issue when rendering single image from image creator.

I agree that we should rewrite `evaluate_filepath_new` to have better check in terms of frame padding of the single frame range.

jakubjezek001:
So here is summary of my findings at original `develop` state of the feature:
1. ``Render Local`` button is creating following problem
image
- so I needed to specify manually the current single frame to be able to render it. 

2. The I've used the ``Read From Rendered`` and it had produced the popup dialogue with ``List index out of range`` . Here is the traceback:
```
  File "<string>", line 1, in <module>
  File "/Applications/Nuke17.0v3/Nuke17.0v3.app/Contents/MacOS/plugins/nukescripts/renderpanel.py", line 7, in render_panel
    return showRenderDialog(_list, exceptOnError, allowFrameServer)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Nuke17.0v3/Nuke17.0v3.app/Contents/MacOS/plugins/nukescripts/renderdialog.py", line 1156, in showRenderDialog
    _showDialog(d)
  File "/Applications/Nuke17.0v3/Nuke17.0v3.app/Contents/MacOS/plugins/nukescripts/renderdialog.py", line 1144, in _showDialog
    dialog.run()
  File "/Applications/Nuke17.0v3/Nuke17.0v3.app/Contents/MacOS/plugins/nukescripts/renderdialog.py", line 550, in run
    nuke.executeMultiple(self._nodeSelection, frame_ranges, views, continueOnError = self._continueOnError.value())
RuntimeError: imageStillFrame.inside_imageStillFrame cannot be executed for multiple frames.
  File "<string>", line 1, in <module>
  File "/Users/jakub/CODE/__YNPUT/ayon-nuke/client/ayon_nuke/startup/write_to_read.py", line 124, in write_to_read
    myfile, firstFrame, lastFrame = evaluate_filepath_new(
                                    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jakub/CODE/__YNPUT/ayon-nuke/client/ayon_nuke/startup/write_to_read.py", line 77, in evaluate_filepath_new
    firstframe = frames[0]
                 ~~~~~~^^^
IndexError: list index out of range
```

This is just for future record. 

Then I was testing latest changes and all works as it is explained.

jakubjezek001 (approved):
works as described

BigRoy:
@rdelillo can you review when you have some time?

client/ayon_nuke/startup/write_to_read.py
```diff
@@ -1,84 +1,106 @@
...
-        # k_value = k_value % first_frame
-        if os.path.isdir(os.path.basename(k_value)):
-            # doesn't check for file, only parent dir
-            filepath = k_value
-        elif os.path.exists(k_eval):
+            os.path.join(project_dir, k_eval)
+        )
+    directory = os.path.dirname(k_eval)
+    if directory and not os.path.isdir(directory):
+        return None, 0, 0
```
rdelillo, line 52: picky, return type becomes ` tuple[str | None, int, int]`  instead of   ` tuple[str, int, int] | None` from docstring.

rdelillo (approved):
Tested latest changes (image + render product), works great now and code reads much better ! Thanks for that 💪

client/ayon_nuke/startup/write_to_read.py
moonyuet: [a861134](https://github.com/ynput/ayon-nuke/pull/363/commits/a86113433f86301ee8e819695756788a342b3729)

PR #364, opened 2026-09-01
Reuse check inventory versions for loader colors YN

jakubjezek001:
adoption of https://github.com/ynput/ayon-nuke/pull/278

## Changelog Description
refactor the tile_color logic to be shared between loaders. Updated the `check_inventory_versions` to use the same logic as the loader for consistent color coding

before:
- each Loader re-implements the `is_latest` checks to update the node colors
- `check_inventory_versions` runs on scriptSave and uses its own logic and own color values to update the node colors
- loading an older version of a product incorrectly used the `latest` color

after:
- `plugin.NukeLoader.get_node_colors` returns a map of "container_status/categories" to colors
   - those values can be overwritten either by setting the respective `node_color_<status>` variables on the LoaderClass
   - or by overwriting `get_node_colors` and return different values
- `plugin.NukeLoader.update_node_color(self, node)` a helper method to update the tile color for a given node. Loaders should use this on `load` and `update`
- `check_inventory_versions` now uses the colors defined by each containers `get_node_colors`

Todo:
- so far I've only updated the `LoadClip` and `LoadImage` loaders as examples. If the general approach is approved I'm happy to roll out the updates to other loaders too.

## Testing notes:
1. import a product using the `LoadClip` or `LoadImage` loader
2. => confirm that the correct color (latest) was used
3. change the version to an older one
4. => confirm that the correct color was used (outdated color)
5. change it back to latest
6. => confirm that the correct color (latest) was used
7. load a outdated version of a product
8. => confirm that the correct color was used (outdated color)
9. save the workfile (this should trigger the `onScriptSave` callback: `check_inventory_versions`
10. confirm that all nodes have the correct colors


BigRoy:
Something went wrong here? This has a massive diff and conflicts that it shouldn't have?

jakubjezek001:
> Something went wrong here? This has a massive diff and conflicts that it shouldn't have?

working on it

jakubjezek001:
@copilot+claude-sonnet-5:medium fix linking and make sure code changes will be minimal required

jakubjezek001:
@BigRoy would you agree, this is better?

BigRoy:
> @BigRoy would you agree, this is better?

Can you describe what you did different than the source PR? Just so I can focus my attention on that.

jakubjezek001:
> > @BigRoy would you agree, this is better?
> 
> Can you describe what you did different than the source PR? Just so I can focus my attention on that.

My intention were resolve the merge conflicts only. The code had been chenged a lot after the #279 in plugins so there were lot of conflicts.

client/ayon_nuke/api/pipeline.py
```diff
@@ -221,6 +221,11 @@ def add_nuke_callbacks(project_settings: dict = None):
 
     nuke_settings = project_settings["nuke"]
 
+    # Set all workfile settings.'
+    nuke.addOnCreate(on_root_create, nodeClass="Root")
+    # set checker for last versions on loaded containers
+    nuke.addOnScriptLoad(check_inventory_versions)
+    # fix ffmpeg settings on script
```
moonyuet, line 228: These lines are duplicated.

moonyuet (commented):
I tested the code and it is working as it is intended. I only have a comment above and it's good to go for approval after it is resolved. Should we support the changes of the color from addon settings?

PR #365, opened 2026-09-11
[POC] generated feature request with use of SDD workflow

jakubjezek001:
## Changelog Description

Adds a dedicated USD camera loader for Nuke with AYON container update support. The new loader loads published USD camera representations, resolves a camera prim automatically, and updates the existing camera node in place when a representation changes. Existing USD camera containers and the legacy UsdCameraLoader contract remain supported.

### What changed

- Added UsdCameraLoaderV2 in client/ayon_nuke/plugins/load/load_camera_usd.py.
- Added specialized matching for product base type camera, representation name usd, and USD extensions usd, usda, and usdc.
- Preserved the existing wildcard UsdCameraLoader for legacy containers and non-camera USD products.
- Added deterministic USD camera-prim discovery, preserving an existing valid prim path during updates.
- Added Camera-family runtime handling with Camera4/Camera3 fallback and validation of required USD knobs.
- Added in-place update, switch, and remove behavior using the existing AYON helpers: containerise, parse_container, and update_container.
- Added rollback of node and container values when an update fails after mutation begins.
- Added the exact UsdCameraLoaderV2 settings key and enabled default under server/settings/loader_plugins.py.
- Guarded the USD Python import so unavailable USD support produces an actionable runtime error instead of an import crash.
- Added SDD documentation under specs/001-nuke-usd-camera-loader/ including the specification, plan, research, data model, contract, tasks, and quickstart.

### Pipeline-contract considerations

- Existing loader identifier UsdCameraLoader is unchanged for saved Nuke containers.
- New containers use the additive identifier UsdCameraLoaderV2.
- Existing legacy loader order remains 2; the new specialized loader uses order 1.
- No publisher, extractor, creator, Pyblish, GUI/menu, startup, or vendor changes are included.
- No settings migration is required because the settings addition is additive.

### Review focus

Please review especially:

1. Whether the new loader compatibility split is appropriate alongside the legacy wildcard loader.
2. Whether in-place update preserves node identity, user knobs, connections, and container metadata.
3. USD prim selection and error/rollback behavior for missing files, invalid prim paths, unreadable stages, and USD files without cameras.
4. Nuke Camera node-version and knob compatibility across supported Nuke versions.
5. Exact-class settings alignment for UsdCameraLoaderV2.

### Automated checks completed

- uv run --with ruff ruff check on the changed Python files — passed.
- uv run --with ruff ruff format --check on the changed Python files — passed.
- uv run python create_package.py --skip-zip — passed.
- git diff --check — passed.

A full repository ruff format check reports formatting issues in unrelated existing files; no unrelated files were reformatted. This repository has no checked-in test suite, so no test framework was added.

### Manual Nuke + AYON validation required

Run in a compatible Nuke + AYON environment:

1. Load a published camera product with a usd representation using UsdCameraLoaderV2.
2. Confirm a USD-capable Camera-family node is created, a camera prim is selected, and the container is listed in Scene Inventory with loader UsdCameraLoaderV2.
3. Rename the node, change a user knob, add a downstream connection, and set a valid custom prim path.
4. Update to a newer representation and verify the same node, name, position, connections, and user edits remain intact.
5. Verify file path, frame rate, prim path, representation id, version metadata, and latest-version color are refreshed.
6. Repeat the update multiple times and confirm metadata does not grow or duplicate.
7. Switch to another version/product and remove the container.
8. Exercise missing-file, unreadable-stage, no-camera-prim, invalid-prior-prim, and unavailable-USD-runtime failure paths; confirm actionable errors and rollback on update failure.
9. Confirm existing Alembic/FBX camera loading, the legacy UsdCameraLoader, and generic USD loaders still work.

***
# Original agentic instrucitons

Here's the prompt — written to get maximum value from the SDD workflow in this repo specifically. I've structured it so the `/speckit.specify` step gets everything it needs to write a compliant spec, and I've baked in the constraints from the constitution, fragments, and this repo's actual code so the workflow doesn't have to rediscover them.

---

## Suggested prompt

> **New feature: USD camera loader for Nuke with update capability**
>
> I want to design and implement a new Loader plugin in the `ayon-nuke` repo
> (`client/ayon_nuke/plugins/load/`) that loads a USD camera into Nuke. It
> must also support **updating** an already-loaded USD camera (the AYON
> "update" workflow — re-resolve a representation and update the container in
> place, preserving node identity/knobs).
>
> **Follow the full Spec-Driven Development workflow** — `/speckit.specify`
> first, then `/speckit.plan`, `/speckit.tasks`, implement with review gates.
> Comply with the AYON addon constitution (`.agents-main/memory/
> ayon-constitution.md`, esp. Articles 2, 3, 6, 7, 8) and the
> `fragments/host-integration.md` guidance. Use the repo's existing helpers
> (`ayon_nuke.api`: `containerise`, `parse_container`, `update_container`,
> `get_node_data`/`set_node_data`) — do not invent parallel metadata formats.
>
> Key requirements to investigate and reflect in the spec:
>
> 1. **Anchor to existing patterns.** Study `client/ayon_nuke/plugins/load/
>    load_camera.py` and `load_camera_usd.py` (if present) plus
>    `load_usd.py`, and the update-capable loaders (e.g. `load_image.py`,
>    `load_model.py`) for the containerise/update_container conventions.
>    Respect the loader-ordering rule — when supported product types or
>    representations overlap, preserve or explicitly adjust ordering.
> 2. **Model & representation contracts.** Confirm the exact AYON product
>    type, `family`/`productBaseType`, and representation naming used for
>    USD cameras (check `ayon-core` loaders and `server/settings/
>    loader_plugins.py`). Identify what `representation_matches`-style
>    filtering should accept.
> 3. **Nuke runtime contracts.** Account for Nuke node class versioning
>    (`Camera3`/`Camera4` — use existing matching logic, don't hard-code),
>    colorspace handling (see `api/colorspace.py`), and knob names as
>    runtime API contracts.
> 4. **Settings contract.** If the loader needs any configuration (paths,
>    options, defaults), add matching settings to `server/settings/
>    loader_plugins.py` (keys = exact plugin class name) and the defaults in
>    `server/settings/main.py`. If it changes any existing settings field
>    names or structure, that needs a `_convert_*` migration in
>    `server/settings/conversion.py` (Article 3). Avoid inventing settings
>    keys not backed by a class name.
> 5. **Runtime boundary (Article 7).** The loader runs inside Nuke (host-API
>    dependent) — keep `addon.py` importable pre-Nuke; GUI-only bits out of
>    headless/farm paths.
> 6. **Vendor/lint (Articles 5/6).** No changes under `client/ayon_nuke/
>    vendor/`; follow this repo's `ruff.toml` (79 cols, 4-space, double
>    quotes).
> 7. **Verification ladder (Article 8).** This repo has no tests — plan
>    `ruff check .`, `ruff format --check .`, `python create_package.py
>    --skip-zip`, plus a clear manual Nuke + AYON validation checklist
>    (create → load → update → publish round-trip). Do not add a test
>    framework.
>
> Open questions I expect the spec to resolve:
> - Does "update" mean re-picking a newer representation of the *same*
>   product (standard container update), or also switching product/version?
> - Any GUI needed (right-click menu entries), or pure loader behavior?
> - Which Nuke node class(es) should the USD camera loader produce, and how
>   should it differ from the existing `load_camera`/`load_camera_usd`
>   loaders (or should it extend/merge them)?
>
> Deliver the spec with the AYON impact-analysis section (settings / host /
> pipeline-contract impact) per the shared `ayon-addon-spec` recipe.

---

## How to run it

In this repo, with the speckit integration initialized (it is — `.specify/` + `.goose/recipes/speckit.*` are present):

1. **`/speckit.specify <the prompt above>`** (or via the `speckit` workflow: `specify → plan → tasks → implement`)
2. Optionally pipe the resulting spec through the shared **`ayon-addon-spec`** recipe (`GOOSE_RECIPE_PATH=<ROOT>/ayon-agentic-instructions/recipes`) to append the AYON impact-analysis section.

---

## Why this prompt shape

- **It front-loads the SDD-critical context** (constitution articles, fragment, existing loader files, settings contract, conversion rule) so the workflow spends its effort on the feature, not on re-deriving repo conventions — and the constitution makes the compliance *checkable* during plan/review.
- **It forces the spec to answer the design forks** (update semantics, node classes, settings keys, ordering) instead of glossing over them — which is where SDD specs usually fail.
- **It names the evidence files** (load_camera.py, load_camera_usd.py, load_usd.py, api/colorspace.py, loader_plugins.py, conversion.py) so the agent reads the *actual* contracts rather than assuming them.
- **It matches this repo's real constraints** (no test framework, vendor excluded, Article 8 ladder, loader ordering) — so the resulting plan/tasks will pass review.

### DIFF
--- a/client/ayon_nuke/plugins/load/load_camera_usd.py
+++ b/client/ayon_nuke/plugins/load/load_camera_usd.py
@@ -1,20 +1,130 @@
+import os
+
 import nuke
 import ayon_api
 
 from ayon_core.pipeline import load
 from ayon_nuke.api import (
     containerise,
+    parse_container,
     update_container,
 )
 from ayon_nuke.api.command import undo_chunk
 from ayon_nuke.api.lib import maintained_selection
 
-from pxr import Usd, UsdGeom
+try:
+    from pxr import Usd, UsdGeom
+except ImportError:  # USD is available only in compatible Nuke runtimes.
+    Usd = None
+    UsdGeom = None
+
+
+def _require_usd():
+    if Usd is None or UsdGeom is None:
+        raise RuntimeError(
+            "USD support is unavailable in this Nuke runtime. "
+            "Use a Nuke build with USD Python support."
+        )
+
+
+def _camera_prim_path(usd_path, existing_path=None):
+    """Return a valid camera prim path from a USD file."""
+    _require_usd()
+
+    if not usd_path or not os.path.exists(usd_path):
+        raise RuntimeError(f"USD camera file does not exist: {usd_path}")
+
+    stage = Usd.Stage.Open(usd_path)
+    if not stage:
+        raise RuntimeError(f"Could not open USD camera file: {usd_path}")
+
+    if existing_path:
+        prim = stage.GetPrimAtPath(existing_path)
+        if prim and prim.IsA(UsdGeom.Camera):
+            return existing_path
+
+    for prim in stage.Traverse():
+        if prim.IsA(UsdGeom.Camera):
+            return prim.GetPath().pathString
+
+    raise RuntimeError(f"No camera prim found in USD file: {usd_path}")
+
+
+def _create_usd_camera_node(object_name, usd_path):
+    """Create the newest available Camera-family USD import node."""
+    last_error = None
+    for node_class in ("Camera4", "Camera3"):
+        try:
+            node = nuke.createNode(
+                node_class,
+                "name {} file {} import_enabled True".format(
+                    object_name, usd_path
+                ),
+                inpanel=False,
+            )
+        except (RuntimeError, NameError) as exc:
+            last_error = exc
+            continue
+
+        required_knobs = (
+            "file",
+            "frame_rate",
+            "import_enabled",
+            "import_prim_path",
+        )
+        missing = [name for name in required_knobs if name not in node.knobs()]
+        if missing:
+            nuke.delete(node)
+            last_error = RuntimeError(
+                f"{node_class} is missing USD knobs: {', '.join(missing)}"
+            )
+            continue
+        return node
+
+    raise RuntimeError(
+        "This Nuke runtime has no USD-capable Camera-family node."
+    ) from last_error
+
+
+def _version_imprint(context):
+    version = context["version"]
+    attributes = version.get("attrib") or {}
+    imprint = {
+        "representation": context["representation"]["id"],
+        "version": version.get("version"),
+        "frameStart": attributes.get("frameStart"),
+        "frameEnd": attributes.get("frameEnd"),
+        "source": attributes.get("source"),
+        "fps": attributes.get("fps"),
+    }
+    return imprint, attributes.get("fps") or nuke.root()["fps"].value()
+
+
+def _node_values(node):
+    values = {}
+    for knob_name in (
+        "file",
+        "frame_rate",
+        "import_enabled",
+        "import_prim_path",
+        "tile_color",
+    ):
+        if knob_name in node.knobs():
+            values[knob_name] = node[knob_name].value()
+    return values
+
+
+def _restore_node_values(node, values):
+    for knob_name, value in values.items():
+        if knob_name in node.knobs():
+            node[knob_name].setValue(value)
 
 
 class UsdCameraLoader(load.LoaderPlugin):
-    """
-    This will load usd camera into script.
+    """Load a USD file containing a camera into Nuke.
+
+    This legacy loader intentionally keeps its broad compatibility contract.
+    Existing containers persist this class name and must continue to resolve.
     """
 
     label = "Load USD Camera"
@@ -23,9 +133,6 @@ class UsdCameraLoader(load.LoaderPlugin):
     order = 2
 
     extensions = {"usd", "usda", "usdc"}
-    # There are essentially no 'camera' product type USD publishers available
-    # in the majority of integrations, so we allow loading any usd
-    # file. This way also USD Shots with cameras can be loaded.
     product_base_types = {"*"}
     product_types = product_base_types
     representations = {"*"}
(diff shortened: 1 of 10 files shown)


client/ayon_nuke/plugins/load/load_camera_usd.py
```diff
@@ -1,20 +1,130 @@
...
+    attributes = version.get("attrib") or {}
+    imprint = {
+        "representation": context["representation"]["id"],
+        "version": version.get("version"),
+        "frameStart": attributes.get("frameStart"),
+        "frameEnd": attributes.get("frameEnd"),
+        "source": attributes.get("source"),
+        "fps": attributes.get("fps"),
+    }
+    return imprint, attributes.get("fps") or nuke.root()["fps"].value()
```
BigRoy, line 100: This seems like odd code, why the FPS as second return value?

Side note: we should not be imprinting fps, source, frameStart, etc. to. That's not default container metadata of AYON. Yes, it matches other areas of Nuke, but really we should never have been imprinting that and should drop that soon if we can.

jakubjezek001: Yes you are right about the imprinting of unnecessary data, its been there for ages.. 

Anyway the attribute from version data is prefered and I believe it might be right, not?

moonyuet:
Neglecting whether the code is a cliché, the quick test shows that the loader can load the usd camera correctly and imprinting the container data correctly.
image

client/ayon_nuke/plugins/load/load_camera_usd.py
BigRoy: > Anyway the attribute from version data is prefered and I believe it might be right, not?

Yes, just saying that the function name and its return data make little sense to me. It's a bit unclear what it's returning really.

jakubjezek001:
closing this, since it was only POC testing

PR #367, opened 2026-09-14
add support for deep write creators

### LINKED ISSUE #25: Nuke Deep Publish

- [X] I have searched the existing issues.

image
Need a deep render/pre-render publisher

### How would you imagine the implementation of the feature?

exactly same approach like write render but with deepwrite

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

_No response_

### Additional context:

_No response_

sebastianbrandhuber:
## Changelog Description
This PR adds Write and Prerender Creator support for DeepWrites. 
Fixes https://github.com/ynput/ayon-nuke/issues/25

- Added creators:
Right now, they are slightly adjusted duplicates of the WriteRender and WritePrerender creator files.
We could also inherit from those classes and just edit what is needed to avoid code duplicates,
but that is open for discussion.
- Added server settings for DeepWriteRender and DeepWritePrerender
- Adjusted plugin and lib to use node_class variable instead of hard-coded "Write" variable.

## Testing notes:
1. Start Nuke, create new creators.


client/ayon_nuke/api/lib.py
```diff
@@ -526,13 +526,14 @@ def get_avalon_knob_data(node, prefix="avalon:", create=True):
     return data
 
 
-def add_write_node(name, file_path, knobs, **kwarg):
+def add_write_node(name, file_path, knobs, node_class, **kwarg):
```
BigRoy: Can we make the default value `="Write"` just for if anyone is using this API elsewhere in some custom script.

sebastianbrandhuber:
@BigRoy @jakubjezek001 @moonyuet 

I was going through all the plugins to see what needs to be adjusted and was wondering if we should change the product (base) types for the deep instances to something like "deeprender" and "deepprerender", and then adjust the codebase (mainly the plugins and their family filters) accordingly -  or leave things as they are?

I think especially parts like [this](https://github.com/ynput/ayon-nuke/blob/d3e30650fa5f6f2cb82c6810467d45d9bca4de38/client/ayon_nuke/plugins/publish/validate_write_nodes.py#L64) could benefit from having different product base types, without requiring too many code changes,
but I currently do not have a clear overview of what impact that would have in general.

I'm open to suggestions or any tips on what you think would be the better approach here!

sebastianbrandhuber:
> @BigRoy @jakubjezek001 @moonyuet
> 
> I was going through all the plugins to see what needs to be adjusted and was wondering if we should change the product (base) types for the deep instances to something like "deeprender" and "deepprerender", and then adjust the codebase (mainly the plugins and their family filters) accordingly - or leave things as they are?
> 
> I think especially parts like [this](https://github.com/ynput/ayon-nuke/blob/d3e30650fa5f6f2cb82c6810467d45d9bca4de38/client/ayon_nuke/plugins/publish/validate_write_nodes.py#L64) could benefit from having different product base types, without requiring too many code changes, but I currently do not have a clear overview of what impact that would have in general.
> 
> I'm open to suggestions or any tips on what you think would be the better approach here!

I went with new product base types to make sure template profiles and filtering stays flexible and can be adjusted
differently for deep products.

client/ayon_nuke/plugins/create/create_deepwrite_prerender.py
```diff
@@ -0,0 +1,87 @@
+import nuke
+
+from ayon_nuke import api as napi
+
+
+class CreateDeepWritePrerender(napi.NukeWriteCreator):
```
moonyuet: This class is pretty much the duplicate of the `CreateWritePrerender`. I think we can just adjust the `create_instance_node` from `CreateWritePrerender` and inherit `CreateDeepWritePrerender` from that class.
Same for `CreateDeepWriteRender` and I think we can inherit it from the existing `CreateWriteRender` class
Some pseudo codes below:
```python
class CreateWritePrerender(napi.NukeWriteCreator):

    settings_category = "nuke"

    identifier = "create_write_prerender"
    label = "Prerender (write)"
    product_base_type = "prerender"
    product_type = product_base_type
    icon = "sign-out"

    instance_attributes = [
        "use_range_limit"
    ]
    default_variants = [
        "Key01",
        "Bg01",
        "Fg01",
        "Branch01",
        "Part01"
    ]

    # Before write node render.
    order = 90

    def create_instance_node(
        self,
        product_name,
        instance_data,
        staging_dir=None,
        node_selection=None,
    ):
        settings = self.project_settings["nuke"]["create"]
        settings = settings[self.__class__.__name__]

        # add fpath_template
        write_data = {
            "creator": self.__class__.__name__,
            "productName": product_name,
            "fpath_template": self.temp_rendering_path_template,
            "staging_dir": staging_dir,
            "render_on_farm": (
                "render_on_farm" in settings["instance_attributes"]
            )
        }

        write_data.update(instance_data)
# ....(More code)

class CreateDeepWritePrerender(CreateWritePrerender):

    identifier = "create_deepwrite_prerender"
    label = "Prerender (deep write)"
    product_base_type = "deepprerender"
    product_type = product_base_type
    node_class = "DeepWrite"


    # Before write node render.
    order = 90

```

BigRoy:
> I was going through all the plugins to see what needs to be adjusted and was wondering if we should change the product (base) types for the deep instances to something like "deeprender" and "deepprerender", and then adjust the codebase (mainly the plugins and their family filters) accordingly - or leave things as they are?

If we can survive **without** requiring a different **base type** that'd be my preference actually. Can we refactor that @sebastianbrandhuber ?

moonyuet:
> @BigRoy @jakubjezek001 @moonyuet
> 
> I was going through all the plugins to see what needs to be adjusted and was wondering if we should change the product (base) types for the deep instances to something like "deeprender" and "deepprerender", and then adjust the codebase (mainly the plugins and their family filters) accordingly - or leave things as they are?
> 
> I think especially parts like [this](https://github.com/ynput/ayon-nuke/blob/d3e30650fa5f6f2cb82c6810467d45d9bca4de38/client/ayon_nuke/plugins/publish/validate_write_nodes.py#L64) could benefit from having different product base types, without requiring too many code changes, but I currently do not have a clear overview of what impact that would have in general.
> 
> I'm open to suggestions or any tips on what you think would be the better approach here!

I think this approach is a very good starting point. My only concern is whether it would hit some odd publishing/transcoding error when we are not using exr as image extension for deepexr writing method (there are `null` or `cdf` in terms of extensions).

client/ayon_nuke/startup/write_to_read.py
```diff
@@ -106,9 +108,12 @@ def detect_file_on_disk(
 
 
 def create_read_node(ndata, comp_start):
-    read = nuke.createNode('Read', 'file "' + ndata['filepath'] + '"')
-    read.knob('colorspace').setValue(int(ndata['colorspace']))
-    read.knob('raw').setValue(ndata['rawdata'])
+    if ndata.get("node_class") == "DeepWrite":
+        read = nuke.createNode('DeepRead', 'file "' + ndata['filepath'] + '"')
+    else:
+        read = nuke.createNode('Read', 'file "' + ndata['filepath'] + '"')
```
iLLiCiTiT:
```suggestion
    filepath = ndata["filepath"]
    if ndata.get("node_class") == "DeepWrite":
        read = nuke.createNode("DeepRead", f'file "{filepath}"')
    else:
        read = nuke.createNode("Read", f'file "{filepath}"')
```

sebastianbrandhuber:
> > I was going through all the plugins to see what needs to be adjusted and was wondering if we should change the product (base) types for the deep instances to something like "deeprender" and "deepprerender", and then adjust the codebase (mainly the plugins and their family filters) accordingly - or leave things as they are?
> 
> If we can survive **without** requiring a different **base type** that'd be my preference actually. Can we refactor that @sebastianbrandhuber ?

@BigRoy I tried that in a first approach and ran into multiple warnings from plugins like the oiio transcode or extract thumbnail plugins,
that use configurations like these:
image

I think that existing bundle configurations like ours (and probably other companies) might not work as expected or without adjustments because different plugins try to handle deep images because they are a **render** or **prerender** product base type.

I can start refactoring the current code, but wanted to make sure beforehand that this is indeed not a problem 
and possible warnings like the examples mentioned above will be handled in those plugins directly?
Not sure how we could exclude deep instances from being processed in other plugins without changing a lot of code....

BigRoy:
> I can start refactoring the current code, but wanted to make sure beforehand that this is indeed not a problem
> and possible warnings like the examples mentioned above will be handled in those plugins directly?
> Not sure how we could exclude deep instances from being processed in other plugins without changing a lot of code....

It's the ongoing debate on how to handle these, but admittedly product base type shouldn't be the solution here. I can see your reasoning from the publishing side of things, but from the AYON resulting data from it it really shouldn't be a different base type.

So I see how it's the road with least friction now, I do still think we'll need to get through it without requiring the base type to be affected.

Even though the base type is used also as publish `family` it doesn't mean that e.g. a deep render can't have an additional `deep` or `render.deep` or similar `family` that publish plug-ins filter too instead, etc.

client/ayon_nuke/plugins/create/create_deepwrite_prerender.py
sebastianbrandhuber: just pushed an update - we could also remove any duplicate class attributes now that e.g. the product base type is the same as in the inherited base class - and strip it down to just the attributes that are different for the new creator.

client/ayon_nuke/plugins/create/create_deepwrite_prerender.py
```diff
@@ -0,0 +1,21 @@
...
+    settings_category = "nuke"
+
+    identifier = "create_deepwrite_prerender"
+    label = "Prerender (deep write)"
+    product_base_type = "prerender"
+    product_type = product_base_type
+    icon = "sign-out"
+    node_class = "DeepWrite"
+
+    instance_attributes = ["use_range_limit"]
```
moonyuet: We can remove the variable of which we haven't done the changes from `CreateWritePrerender`
```suggestion
    identifier = "create_deepwrite_prerender"
    label = "Prerender (deep write)"
    node_class = "DeepWrite"

```

client/ayon_nuke/plugins/create/create_deepwrite_render.py
```diff
@@ -0,0 +1,16 @@
...
+
+class CreateDeepWriteRender(CreateWriteRender):
+    settings_category = "nuke"
+
+    identifier = "create_deepwrite_render"
+    label = "Render (deep write)"
+    product_base_type = "render"
+    product_type = product_base_type
+    icon = "sign-out"
+    node_class = "DeepWrite"
```
moonyuet, line 9: We can remove the variable of which we haven't done the changes from CreateWriteRender
```suggestion
    identifier = "create_deepwrite_render"
    label = "Render (deep write)"
    node_class = "DeepWrite"
```

client/ayon_nuke/plugins/create/create_deepwrite_prerender.py
moonyuet: > just pushed an update - we could also remove any duplicate class attributes now that e.g. the product base type is the same as in the inherited base class - and strip it down to just the attributes that are different for the new creator.

Agreed - I have already commented suggestions for the removal

sebastianbrandhuber:
With the current state of the code and going with "render" / "prerender" as a product base type,
I get an Error for the extract thumbnail plugin:

```
2026/09/15 14:10:33 WARNING: Failed to create thumbnail using oiiotool
  File "C:\Users\brandhuber\AppData\Local\Ynput\AYON\addons\core_1.9.12\ayon_core\plugins\publish\extract_thumbnail.py", line 593, in _create_thumbnail_oiio
    run_subprocess(oiio_cmd, logger=self.log)
  File "C:\Users\brandhuber\AppData\Local\Ynput\AYON\addons\core_1.9.12\ayon_core\lib\execute.py", line 182, in run_subprocess
    raise RuntimeError(exc_msg)
RuntimeError: Executing arguments was not successful: "(['C:\\Users\\brandhuber\\AppData\\Local\\Ynput\\AYON\\addons_resources\\ayon_third_party\\oiio_windows_83e412e9\\bin\\oiiotool.exe', '-i:ch=Z,A', '\\\\blackhole\\media\\vfx\\hellotobias_11144\\0001_010\\comp\\work\\0001_010_comp_compDeepMain_sbr\\0001_010_comp_compDeepMain_sbr.1033.exr', '--ch', 'R=Z,G=Z,B=Z', '--subimage', '0', '-o', 'C:\\Users\\BRANDH~1\\AppData\\Local\\Temp\\pyblish_tmp_js_fkn8z\\0001_010_comp_compDeepMain_sbr.1033_thumb.jpg'],)"Error:
oiiotool ERROR: -o : jpeg does not support 'deep' images
```
But the publish seems to be successfull

client/ayon_nuke/plugins/publish/validate_exposed_knobs.py
```diff
@@ -55,19 +55,22 @@ class ValidateExposedKnobs(
 
     settings_category = "nuke"
 
-    product_base_types_mapping = {
-        "render": "CreateWriteRender",
-        "prerender": "CreateWritePrerender",
-        "image": "CreateWriteImage"
+    creator_identifier_mapping = {
```
sebastianbrandhuber: this part still feels very rigid.

what about getting the plugin name from the create context?

```
creator_identifier = instance.data["creator_identifier"]
creator = instance.context.data["create_context"].creators.get(
    creator_identifier
)
plugin_name = creator.__class__.__name__
```

same for the validate_write_nodes plugin

client/ayon_nuke/api/plugin.py
```diff
@@ -596,6 +602,10 @@ def get_instance_group_node_children(instance):
 
 
 def get_colorspace_from_node(node):
+    # DeepWrite nodes do not have a colorspace, so return None.
+    if node.Class() == "DeepWrite":
+        return None
```
BigRoy, line 607: Doesn't this change the return type for this function from `str` to `str | None`? Not sure if we're super happy with that? Which means that on extract for these nodes we may just start guessing the colorspace on the filename?

Also, why wouldn't colors in a deep EXR have a colorspace? Wouldn't/Shoukdn't it still have a colorspace somewhere? I mean, I can see that DeepWrite node doesn't have the colorspace attribute - but how does one then actually color manage deep EXRs? 🤔 (This is probably a stupid question?)

@jakubjezek001 do you know?

client/ayon_nuke/plugins/publish/validate_exposed_knobs.py
```diff
@@ -55,19 +59,19 @@ class ValidateExposedKnobs(
...
 
-        product_base_type = instance.data["productBaseType"]
-        plugin = self.product_base_types_mapping[product_base_type]
         group_node = instance.data["transientData"]["node"]
+
+        creator_identifier = instance.data["creator_identifier"]
+        creator = instance.context.data["create_context"].creators.get(
+            creator_identifier
+        )
+        plugin = creator.__class__.__name__
```
BigRoy, line 73: We do this about three times here - should we put a helper function for this in `plugin.py`

```python
get_creator_class_name(instance: pyblish.api.Instance) -> str:
    """Return creator class name from publish instance.
    
    Often used as lookup in settings for creator-specific overrides.
    """
    creator_identifier = instance.data["creator_identifier"]
    creator = instance.context.data["create_context"].creators.get(
        creator_identifier
    )
    return creator.__class__.__name__
```

moonyuet:
> With the current state of the code and going with "render" / "prerender" as a product base type, I get an Error for the extract thumbnail plugin:
> 
> ```
> 2026/09/15 14:10:33 WARNING: Failed to create thumbnail using oiiotool
>   File "C:\Users\brandhuber\AppData\Local\Ynput\AYON\addons\core_1.9.12\ayon_core\plugins\publish\extract_thumbnail.py", line 593, in _create_thumbnail_oiio
>     run_subprocess(oiio_cmd, logger=self.log)
>   File "C:\Users\brandhuber\AppData\Local\Ynput\AYON\addons\core_1.9.12\ayon_core\lib\execute.py", line 182, in run_subprocess
>     raise RuntimeError(exc_msg)
> RuntimeError: Executing arguments was not successful: "(['C:\\Users\\brandhuber\\AppData\\Local\\Ynput\\AYON\\addons_resources\\ayon_third_party\\oiio_windows_83e412e9\\bin\\oiiotool.exe', '-i:ch=Z,A', '\\\\blackhole\\media\\vfx\\hellotobias_11144\\0001_010\\comp\\work\\0001_010_comp_compDeepMain_sbr\\0001_010_comp_compDeepMain_sbr.1033.exr', '--ch', 'R=Z,G=Z,B=Z', '--subimage', '0', '-o', 'C:\\Users\\BRANDH~1\\AppData\\Local\\Temp\\pyblish_tmp_js_fkn8z\\0001_010_comp_compDeepMain_sbr.1033_thumb.jpg'],)"Error:
> oiiotool ERROR: -o : jpeg does not support 'deep' images
> ```
> 
> But the publish seems to be successfull

you might need to pass the `--flatten` flag to extract thumbnail to make it successful

BigRoy (commented):
Looking pretty good to me - did have some questions

client/ayon_nuke/api/plugin.py
moonyuet: image

You can see the difference between the two write nodes.

BigRoy: Yes, I checked as well - it doesn't have it. My question was merely a usage question - if the write node doesn't have it, how do you actually manage colors for Deep EXRs in your production? Wouldn't you still need OCIO somewhere at some point? 🤔

sebastianbrandhuber: As far as I know, deep can contain raw rgb values next to depth samples,
but not in a traditionally "color-managed" way.

Another possible problem: If we return None,
the colorspace might also be wrongly assigned from filepath rules:
https://github.com/ynput/ayon-core/blob/f702eb95551b86f3cb4963de924a57b819ffcd2c/client/ayon_core/pipeline/colorspace.py#L1317

If we want to set a colorspace, we probably need to come up with a solution to get the RAW colorspace?
Maybe with the data role, or by guessing, but that would mean that we need a color-managed environment
and have to resolve it by a given publish context.

jakubjezek001: My assumption is mostly that DeepData exr should be usually role `data` which is RAW. 

It would only make sense to assign colorspaceData if we wanted to do some conversions to it, but I would rather not be doing so and keep data as they are.

server/settings/imageio.py
```diff
@@ -367,6 +373,30 @@ class ImageIOSettings(BaseSettingsModel):
...
+                    {"type": "text", "name": "file_type", "text": "exr"},
+                    {
+                        "type": "text",
+                        "name": "datatype",
+                        "text": "16 bit half",
+                    },
+                    {
+                        "type": "color_gui",
+                        "name": "tile_color",
+                        "color_gui": [186, 35, 35],
```
jakubjezek001, line 390:
```suggestion
                        "color_gui": [0, 3, 92],
```
I would keep the original Deep colorspace at least on Render product type

Image

jakubjezek001 (changes requested):
the colorspace change is optional only recommended. 

I was testing for now only Create and Publish on Render product type.

Validation stopped me. `Missing Reformat node in render group node` - either add exception in the validator for skipping Deep related instances or we need to add the Reformat node. 

***
and thanks a lot for your submission!

PR #371, opened 2026-09-16
Check outdated containers with popup on scene load

sebastianbrandhuber:
## Changelog Description
This Draft PR implements a check for outdated containers **OnScriptLoad**, very similar to houdini.

For now, I leave the following points open for discussion:

- Should we add a server setting in the Workfile Callbacks so that this is configurable?
- Do we also want the check to happen **OnScriptSave**? For now, 
the check with a popup only happens **OnScriptLoad** and coloring the nodes happens on both **OnScriptLoad** and **OnScriptSave**.
- I added a **nuke.GUI** check in lib.check_inventory_versions,
but I think we could also implement this way earlier in pipeline.on_script_load because it should not happen without a GUI

## Testing notes:
1. Open a Nuke file that has outdated containers.
2. The popup should show in the lower right corner like in Houdini.


BigRoy (approved):
Logic looks ok to me.

> Should we add a server setting in the Workfile Callbacks so that this is configurable?

I don't think it's needed - also doesn't exist in other DCCs.

> Do we also want the check to happen OnScriptSave? For now, the check with a popup only happens OnScriptLoad and coloring the nodes happens on both OnScriptLoad and OnScriptSave.

It should behave like other DCCs I'd say - in e.g. Maya we only do so on open as well. Fine with me as such.

> I added a nuke.GUI check in lib.check_inventory_versions,
but I think we could also implement this way earlier in pipeline.on_script_load because it should not happen without a GUI

I'd actually argue that this check should not be within this method, but would be responsibility of the caller instead. At the same time, it's not a big problem the way it is now.

client/ayon_nuke/api/lib.py
```diff
@@ -832,12 +833,24 @@ def get_view_process_node():
...
+    if not nuke.GUI:
+        return
+
+    if show_popup:
+        try:
+            if any_outdated_containers():
+                log.warning("Scene has outdated content.")
+                _show_outdated_content_popup()
+        except Exception as error:
+            log.warning(error)
```
BigRoy: Let's maintain the stack trace.
```suggestion
            log.warning(error, exc_info=True)
```

moonyuet (approved):
Code looks good for me and tested successfully in 16.0

